PR #15, opened 2020-06-05
Change openssh to SSHListener format and add port 2222 for local ssh.

Cryptiiiic:
- Add sshd-keygen-wrapper
- Make daemon use sockets
- Port 2222 local ssh
- ~~Adjust postinst~~ Remove postinst


build_info/openssh.postinst
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+[ ! -f /etc/ssh/ssh_host_key ]     && ssh-keygen -q -t rsa1 -f /etc/ssh/ssh_host_key     -N "" -C "" < /dev/null > /dev/null 2> /dev/null
```
Diatrus, line 2: This postinst shouldn't be needed anymore with that libexec wrapper. Remove it and this should be fine.

PR #20, opened 2020-06-06
Fixes tmux issue #19.

### LINKED ISSUE #19: tmux is missing the entitlements xml file.
tmux.mk references tmux.xml in the sign stage but tmux.xml does not exist.
tmux gets killed because the sign stage failed leading to tmux not having any entitlements.

Cryptiiiic:
This fixes #19.


Diatrus:
Why is top.xml needed? Does tmux really need task-for-pid-allow? I feel like general.xml would be more suited to this.

PR #22, opened 2020-06-08
Add libzip & libfragmentzip

M1staAwesome:


libfragmentzip.mk
```diff
@@ -0,0 +1,44 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS  += libfragmentzip
```
Diatrus, line 5: Could you format these correctly?
All `=` should be vertically parallel directly above/below another one.

libzip.mk
```diff
@@ -0,0 +1,56 @@
...
+		-DCMAKE_CROSSCOMPILING=true \
+		-DCMAKE_INSTALL_NAME_TOOL=$(I_N_T) \
+		-DCMAKE_INSTALL_PREFIX=/ \
+		-DCMAKE_INSTALL_NAME_DIR=/usr/lib \
+		-DCMAKE_INSTALL_RPATH=/usr \
+		-DCMAKE_OSX_SYSROOT="$(SYSROOT)" \
+		-DCMAKE_C_FLAGS="$(CFLAGS)" \
+		-DCMAKE_CXX_FLAGS="$(CXXFLAGS)" \
+		-DCMAKE_SHARED_LINKER_FLAGS="-L$(BUILD_BASE)/usr/lib" \
+		-DCURRENT_VENDOR=debian \
```
Diatrus, line 30: You likely don't need current vendor or any options like that; That's an apt specific option. Vet these options and deduce which ones aren't necessary.

libfragmentzip.mk
M1staAwesome: thought i fixed that with [c65bcf6](https://github.com/ProcursusTeam/Procursus/pull/22/commits/c65bcf63e76168444b64e2be72d87202a969284a), but must've undone it, weird

PR #28, opened 2020-06-15
Download source code packages from GNU mirror where possible

M1staAwesome:


Diatrus:
Is there a reason for this?

M1staAwesome:
No particular reason, just (theoretical) faster response times from mirrors.

PR #32, opened 2020-06-16
Add tor

CRKatri:


tor.mk
```diff
@@ -0,0 +1,50 @@
...
+DEB_TOR_V    ?= $(TOR_VERSION)
+
+tor-setup: setup
+	$(call EXTRACT_TAR,tor-$(TOR_VERSION).tar.gz,tor-$(TOR_VERSION),tor)
+
+ifneq ($(wildcard $(BUILD_WORK)/tor/.build_complete),)
+tor:
+	@echo "Using previously built tor."
+else
+tor: tor-setup
```
Diatrus, line 17: Does tor link to openssl, libevent, xz, and/or zstd? If so, why aren't they a required target here, and if not, why are they listed as Depends in the control file?

tor.mk
```diff
@@ -0,0 +1,50 @@
...
+		LFLAGS2="$(CFLAGS)"
+	+$(MAKE) -C $(BUILD_WORK)/tor install
+	touch $(BUILD_WORK)/tor/.build_complete
+endif
+
+tor-package: tor-stage
+	# tor.mk Package Structure
+	rm -rf $(BUILD_DIST)/tor
+	mkdir -p $(BUILD_DIST)/tor
+	mkdir -p $(BUILD_DIST)/tor/usr
```
Diatrus, line 34: The first mkdir command is useless. Just include `mkdir -p $(BUILD_DIST)/tor/usr` instead of both lines.

tor.mk
```diff
@@ -0,0 +1,50 @@
...
+
+tor-setup: setup
+	$(call EXTRACT_TAR,tor-$(TOR_VERSION).tar.gz,tor-$(TOR_VERSION),tor)
+
+ifneq ($(wildcard $(BUILD_WORK)/tor/.build_complete),)
+tor:
+	@echo "Using previously built tor."
+else
+tor: tor-setup
+	cd $(BUILD_WORK)/tor && ./configure --prefix=$(BUILD_STAGE)/tor \
```
Diatrus, line 18: I have no doubt specifying `--prefix=/usr` would alleviate the issue you have during the package stage with everything installing to `$(BUILD_STAGE)/bin` and `$(BUILD_STAGE)/share`.

CRKatri: Setting the prefix to /usr will move /etc/tor to /usr/etc/tor I’ll look into how to fix that tonight

tor.mk
```diff
@@ -30,13 +34,12 @@ endif
 tor-package: tor-stage
 	# tor.mk Package Structure
 	rm -rf $(BUILD_DIST)/tor
-	mkdir -p $(BUILD_DIST)/tor
+	mkdir -p $(BUILD_DIST)/tor/etc
 	mkdir -p $(BUILD_DIST)/tor/usr
```
Diatrus, line 38: You can get rid of all of these and just `cp -a $(BUILD_STAGE)/tor $(BUILD_DIST)`

Diatrus:
Thanks for the contribution!

PR #35, opened 2020-06-21
fix groff detection when using brew

EduardoRFS:
## Problem
On macOS it is using `/usr/bin/groff` instead of brew groff

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -328,6 +328,9 @@ $(error Install flex)
 endif
 
 ifneq ($(call HAS_COMMAND,groff),1)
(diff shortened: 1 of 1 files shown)


Diatrus:
/usr/local/bin is pointedly avoided in every case because some people have other tools that pointedly conflict with procursus installed there. I would recommend adding /usr/local/opt/groff/bin or the equivalent to PATH instead, where it detects groff. This is a recent oversight on my part.

Diatrus:
Thank you for the contribution. It makes me really happy when others notice my mistakes, heh.

PR #39, opened 2020-06-26
add whois and libcrypt1

1Conan:


build_info/whois.control
```diff
@@ -0,0 +1,12 @@
+Package: whois
+Maintainer: @DEB_MAINTAINER@
+Architecture: @DEB_ARCH@
+Version: @DEB_WHOIS_V@
+Depends: libidn2, libcrypt1
```
Diatrus, line 5: As far as I remember libcrypt1 isn't a real dependency ;)

whois.mk
```diff
@@ -0,0 +1,50 @@
...
+	wget -q -nc -P $(BUILD_SOURCE) http://deb.debian.org/debian/pool/main/w/whois/whois_$(WHOIS_VERSION).tar.xz
+	$(call EXTRACT_TAR,whois_$(WHOIS_VERSION).tar.xz,whois-$(WHOIS_VERSION),whois)
+	$(SED) -i '/_XOPEN_SOURCE/d' $(BUILD_WORK)/whois/utils.c 
+	$(SED) -i '/_XOPEN_SOURCE/d' $(BUILD_WORK)/whois/mkpasswd.c 
+
+ifneq ($(wildcard $(BUILD_WORK)/whois/.build_complete),)
+whois:
+	@echo "Using previously built whois."
+else
+whois: whois-setup libidn2 libcrypt1
```
Diatrus, line 19: Same comment as above

whois.mk
```diff
@@ -0,0 +1,50 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+STRAPPROJECTS  += whois
```
Diatrus, line 5: This should be SUBPROJECTS, not STRAPPROJECTS. (Affects bootstrap.tar creation)

libcrypt1.mk
```diff
@@ -0,0 +1,47 @@
...
+
+libcrypt1-setup: setup
+	wget -q -nc -P $(BUILD_SOURCE) https://deb.debian.org/debian/pool/main/libx/libxcrypt/libxcrypt_$(LIBCRYPT1_VERSION).orig.tar.xz
+	$(call EXTRACT_TAR,libxcrypt_$(LIBCRYPT1_VERSION).orig.tar.xz,libxcrypt-$(LIBCRYPT1_VERSION),libcrypt1)
+
+ifneq ($(wildcard $(BUILD_WORK)/libcrypt1/.build_complete),)
+libcrypt1:
+	@echo "Using previously built libcrypt1."
+else
+libcrypt1: libcrypt1-setup gettext
```
Diatrus, line 16: If gettext is actually linked to libcrypt, it should be a depend in the control file as well! :)

build_info/whois.control
1Conan: compiling mkpasswd as well to be inline with debian

libcrypt1 is needed for that.

Diatrus: Is mkpasswd useful on a Darwin os? Have you tested it performs its intended function?

1Conan: I dunno if it's useful for anyone but I just included it in case someone expect it to be available in the whois package just like debian.

Diatrus: I didn’t mean useful in the sense someone would have an interest in using it— I was more asking if it was even functional.

1Conan: Oh oops I forgot to mention that, yes it is.

PR #42, opened 2020-06-27
Add nanorc

M1staAwesome:
#makenanohotagain

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "apt"]
 	path = apt
 	url = https://github.com/Diatrus/apt.git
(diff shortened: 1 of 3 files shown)


Diatrus:
There are a few reasons I'm hesitant to approve this:
* Rather than this being a tool, it seems to be a large set of configuration files.
* It offers a way to easily install through wget/curl.
* Configuring for root and mobile in a postinst is undesirable, especially since it overwrites previous nanorcs the user may have configured.
I'm essentially just confused why this has to be a package.

M1staAwesome:
It was mostly for convenience, but I can understand why you wouldn't want it added.

PR #43, opened 2020-06-28
tsschecker and img4tool

1Conan:


curl.mk
```diff
@@ -15,7 +15,7 @@ ifneq ($(wildcard $(BUILD_WORK)/curl/.build_complete),)
 curl:
 	@echo "Using previously built curl."
 else
-curl: curl-setup openssl libssh2 nghttp2 libidn2
+curl: curl-setup openssl libssh2 nghttp2 libidn2 libunistring
```
Diatrus, line 18: When adding new dependencies to a package, please make sure to add the dependency to its respective control file.

libzip.mk
```diff
@@ -14,7 +14,7 @@ ifneq ($(wildcard $(BUILD_WORK)/libzip/.build_complete),)
 libzip:
 	@echo "Using previously built libzip."
 else
-libzip: libzip-setup
+libzip: libzip-setup xz
```
Diatrus, line 17: Same comment as curl.

curl.mk
1Conan: forgot about those again kek

PR #44, opened 2020-06-29
libtool 2.4.6

M1staAwesome:
Program is prefixed with a `g` as per Brew.


build_info/libtool.control
```diff
@@ -0,0 +1,7 @@
+Package: libtool
+Version: @DEB_LIBTOOL_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section: Development
+Priority: optional
+Description: GNU Libtool is a generic library support script. Libtool hides the complexity of using shared libraries behind a consistent, portable interface.
```
Diatrus, line 7: Could you add `Name: GNU libtool` here to make sure nobody misunderstands that this is **mostly** unnecessary.

Diatrus: Also remember to leave a newline

Diatrus: A disclamer at the end of the description saying "This libtool is prefixed with `g` in order to not conflict with Darwin CC Tools." would be nice as well.

build_info/libtool.control
```diff
@@ -0,0 +1,7 @@
+Package: GNU libtool
```
Diatrus, line 1: This would be libtool. `Name:` is a completely separate entry in the control file.

PR #45, opened 2020-06-29
Rest of autotools

M1staAwesome:


demhademha:
Do you think you could add  autossh and pkg-config?

PR #48, opened 2020-07-01
Fix my stupid mistake with tor

CRKatri:


tor.mk
```diff
@@ -15,10 +15,10 @@ tor:
 	@echo "Using previously built tor."
 else
 tor: tor-setup libevent openssl xz zstd
-	cd $(BUILD_WORK)/tor && ./configure --prefix=$(BUILD_STAGE)/tor/usr \
+	cd $(BUILD_WORK)/tor && ./configure --prefix=/usr \
```
Diatrus, line 18: Put prefix on a newline please

PR #52, opened 2020-07-04
Rename SYSROOT to TARGET_SYSROOT

CRKatri:
This allows declaring a custom sysroot path in a .zshrc or .bashrc without breaking the compilation of other packages. This should have hit every instance as I did a find and replace using VS Code in all files. Please let me know if I missed any and I'll fix it right away.


Makefile
```diff
@@ -102,7 +102,7 @@ ARCH := $(shell echo $(ARCHES) | awk -F' ' '{ for(i=1;i<=NF;i++) print "-arch "
 
 ifeq ($(UNAME),Linux)
 $(warning Building on Linux)
-SYSROOT         ?= $(HOME)/cctools/SDK/iPhoneOS13.2.sdk
+TARGET_SYSROOT         ?= $(HOME)/cctools/SDK/iPhoneOS13.2.sdk
```
Diatrus, line 105: Please fix the formatting here. See the above line. You’ll know what I mean. Same with any other instance of this happening.

PR #57, opened 2020-07-11
Enable UTF8 support in nano + Fix reading system config from /etc/nanorc

marijuanARM:


Diatrus:
Does enabling utf8 create a new dependency for nano that you didn't account for here?

marijuanARM:
Just tried with no other packages compiled, doesn't look like it does.

marijuanARM:
Also noticed from reading through nano documentation, looks like nano will read the system nanorc from /usr/bin/etc/nanorc, so a symlink might be needed
Relevant

Diatrus:
Did you test this on device? If not, can you? If so, did it read from /etc/nanorc properly?

marijuanARM:
Yup, reads from /etc/nanorc perfectly (and made sure that it wasn't just a ~/.nanorc being read that made me think it was reading /etc/nanorc).

PR #65, opened 2020-07-24
Add jq 1.6 and oniguruma 5.9.4

absidue:
oniguruma is a regex library required by jq


Diatrus:
If it isn't too much trouble, would you do be the huge favor of debianizing these packages before I merge?

An example of how it's done can be found [here](https://github.com/ProcursusTeam/Procursus/blob/eef935e2548b90294a811b2703ed2ac075beb5e0/libplist.mk#L31).

The best way to find what should be in each package is to look at the .install files in each Debian git repo. [JQ's, for example](https://salsa.debian.org/debian/jq/-/tree/master/debian). If you need any assistance, just DM me on Discord.

PR #67, opened 2020-07-28
Add pfetch

CRKatri:
Current says the OS is macOS however, a [PR has been opened](https://github.com/dylanaraps/pfetch/pull/67) to fix that on upstream so it will say macOS until that is merged, or some other fix is made.


Diatrus:
Any reason this is a submodule?

CRKatri:
It rarely has releases, however I will switch it to the latest release.

PR #68, opened 2020-07-30
Add tree

CRKatri:
The way the Makefile was written it doesn't support `DESTDIR` so I have to use prefix which shouldn't have any problems. Patching the Makefile is the official method to compile.


Diatrus:
Thank you so much for the continued contributions!

PR #74, opened 2020-08-04
Allow the destination in EXCTRACT_TAR to be the same as the directory to extract

absidue:


Diatrus:
I believe what you're looking for is already covered by an undocumented option. Adding a `,1` on the end of the EXTRACT_TAR call should do the trick. Have a look at the perl makefile for example.

absidue:
Okay thanks

PR #75, opened 2020-08-04
Fix Installed-Size showing in transmission description & Add some conflicts to dimentio

marijuanARM:


Diatrus:
Thanks for fixing my oversight :)

PR #77, opened 2020-08-05
Update berkleydb to 18.1.40

CRKatri:
The version update has caused the old version to be removed from the mirror (or just moved) so updating the version is required to download the source code.


Diatrus:
Please don’t use any oracle links.

PR #78, opened 2020-08-08
Added fff (F---ing Fast File-Manager) program

zamiur:
Version of this program is 2.1.


CRKatri:
This is missing a control file.

zamiur:
Hi, thanks for pointing that out. I'll add one :P

CRKatri:
It needs to be called fff.control

zamiur:
I know, It filled it as .mk instead

PR #79, opened 2020-08-08
Add mpc and mpfr

brennanleblanc0:
Complex floating-point and multiple-precision floating-point libraries respectively.

mpc: http://www.multiprecision.org/mpc/
mpfr: https://www.mpfr.org/

From my testing, setup, building and packaging work for both.


mpc.mk
```diff
@@ -0,0 +1,44 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS  += mpc
+MPC_VERSION := 1.1.0
+DEB_MPC_V   ?= $(MPC_VERSION)
```
CRKatri, line 7: Fix the formating here.

brennanleblanc0:
yeah needs an overhaul

PR #80, opened 2020-08-08
Add mpc and mpfr

brennanleblanc0:
Complex floating-point and multiple-precision floating-point libraries respectively.

mpc: http://www.multiprecision.org/mpc/
mpfr: https://www.mpfr.org/

From my testing, setup, building and packaging work for both.

Fixes many issues with #79


Diatrus:
Looks good — curious what these are used for?

brennanleblanc0:
They are used as dependencies for GCC, which I took a break on trying. These were already made and working so I decided submit. I'm sure some people/devs may find a purpose for it

PR #81, opened 2020-08-08
Fix the logging while packaging lolcat (lolcat != rsync)

absidue:


CRKatri:
Whoops, sorry about that.

Diatrus:
Lmao classic case of copy pasting makefiles. I’m guilty as well.

PR #85, opened 2020-08-09
Add lynx

CRKatri:


CRKatri:
Just noticed I forgot to change `--with-screen=` to `ncursesw` in this PR, that fix should be applied. sorry about that.

PR #102, opened 2020-08-12
Add pacman and libalpm

CRKatri:


Diatrus:
Can you explain where pacman would come in use on the repo?

CRKatri:
Pacman is required to use `makepkg`.

Diatrus:
Ah. I presume makepkg is included in pacman?

CRKatri:
Yes.

Diatrus:
Could you do me a favor and rename pacman's package to makepkg? I want it to be clear that this is a utility for arch users rather than an apt-alternative.

PR #145, opened 2020-10-02
Update tor

CRKatri:
Doesn't have a LaunchDaemon yet, I do not know why it refused to work, so it's on hold.


Diatrus (approved):
Thanks lots

PR #164, opened 2020-10-10
Fixed libtool

demhademha:


CRKatri:
This PR actually breaks libtool, I also do not see the issue that this is supposed to solve.

PR #166, opened 2020-10-12
Add zsign 

demhademha:
Thanks to @1Conan  for their support


Diatrus:
Please version it `0~(DATEOFLASTCOMMIT)`

For example, if the date of the last commit was 10/12/2020 the version would be 0~20201012

demhademha:
This is fixed now

PR #172, opened 2020-10-20
Add GitHub CLI

marijuanARM:
wrote & tested completely on my phone :p

### DIFF
--- a/build_info/libedit-dev.control
+++ b/build_info/libedit-dev.control
@@ -1,0 +1,12 @@
+Package: libedit-dev
+Version: @DEB_LIBEDIT_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section: Development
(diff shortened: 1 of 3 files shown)


gh.mk
```diff
@@ -0,0 +1,47 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS += gh
+GH_VERSION  := 1.1.0
+DEB_GH_V    ?= $(GH_VERSION)
+
+gh-setup: setup
+	wget -q -nc -P $(BUILD_SOURCE) https://github.com/cli/cli/archive/v$(GH_VERSION).tar.gz
```
Diatrus, line 10: Please don't download with only version identifier as the tar. Look at brotli.mk for an example on how to do it properly in this case.

Diatrus (changes requested):

PR #182, opened 2020-10-31
Fix speedtest-cli (sorry about that)

CRKatri:


Diatrus (changes requested):
Need to account for this change in the control file :)

PR #186, opened 2020-10-31
Add /etc/shells

CRKatri:


Diatrus:
What does this solve that we don’t already have?

CRKatri:
Currently we don't have a `/etc/shells` which can be required for some programs, like dropbear which it checks you login shell against, so this adds the `add-shells` and `remove-shells` scripts in debianutils and sets it up for the shells we have currently

lafi8877 (approved):

PR #209, opened 2020-11-23
Add more detailed Linux install instructions

Randomblock1:
Should get you from 0-100 on Linux real quick.
Includes step-by-step instructions for every part of setup.

### DIFF
--- a/LINUX.md
+++ b/LINUX.md
@@ -10,7 +10,7 @@ Here is how, exactly, to install cctools and use it to compile.
 - You're going to need an iOS 13 SDK. Since you're on Linux, you're going to have to search online for one. I personally use https://github.com/xybp888/iOS-SDKs.
 
 3. Compress SDK
-- Compress iPhoneOS.sdk to iPhoneOS.sdk.tar.xz. On Ubuntu you can use the Files program and compress files by right-clicking and selecting Compress. 
+- Compress iPhoneOS.sdk to iPhoneOS.sdk.tar.xz. ```tar caf iPhoneOS.sdk.tar.xz iPhoneOS.sdk```
 
 4. Get CCTools Port
 - You need Git for this. ```git clone https://github.com/tpoechtrager/cctools-port```
--- a/LINUX.md
+++ b/LINUX.md
@@ -66,12 +66,12 @@ rm MacOSX10.15.sdk.tar.xz
 
 You can now compile Procursus packages!
 
-If you used a different SDK, you need to set TARGET_SYSROOT=$(HOME)/cctools/SDK/YOUR_SDK_HERE.
+If you used a different SDK, you need to set `TARGET_SYSROOT=$(HOME)/cctools/SDK/YOUR_SDK_HERE`.
 Ideally, you should set it inside your .profile, .bashrc, or .zshrc. Then, open a new shell or use ```source``` again.
 
 Test your toolchain by trying to build bash: 
 
-```make bash NO_GPG=1```
+```make bash```
 
 This will build bash quickly, so you can test if your toolchain works. If it does, congratulations! 
 You should now be able to compile almost everything in Procursus with Linux. 
--- a/LINUX.md
+++ b/LINUX.md
@@ -1,7 +1,9 @@
 # Building on Linux
 Building on linux is made possible by cctools-port.
 
-Here is how, exactly, to install cctools and use it to compile.
+### The easiest way to do this is by using [this script](https://gist.github.com/1Conan/4347fd5f604cfe6116f7acb0237ef155).
+
+If you want to do it manually, however, follow these instructions.
 
 1. Dependencies
 - You need: clang (>=3.4) and everything Procursus normally needs. Get clang with apt.
--- a/LINUX.md
+++ b/LINUX.md
@@ -9,22 +9,37 @@ If you want to do it manually, however, follow these instructions.
 - You need: clang (>=3.4) and everything Procursus normally needs. Get clang with apt.
 
 2. Get iOS SDK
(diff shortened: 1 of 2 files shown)


LINUX.md
```diff
@@ -0,0 +1,46 @@
...
+```
+- And finally, ```source ~/.profile```. You'll need to re-source every time you use a new shell.
+
+7. Get MacOS SDK
+
+Download the MacOS 10.15 SDK from https://github.com/phracker/MacOSX-SDKs and extract it, then move it to ~/cctools/SDK/MacOSX.sdk.
+
+You can now compile Procursus... with some tweaking. You'll need to set some options manually with your custom build.
+
+You need to set GNU_HOST_TRIPLE=aarch64-apple-darwin11 or whatever cctools generated for you (check the names of files in ~/cctools/bin).
```
CRKatri, line 39: DO NOT modify GNU_HOST_TRIPLE, the value of that must be set by Procursus, or nothing will build correctly

LINUX.md
```diff
@@ -0,0 +1,46 @@
...
+- The last step made a toolchain in the "target" folder. We can use this to compile Procursus, but first we need to move it to our home folder and tell our system to use it.
+- ```mv target ~/cctools```
+- Then add this to your .profile:
+
+```
+if [ -d "$HOME/cctools" ] ; then
+    PATH="$HOME/cctools/bin:$PATH"
+fi
+```
+- And finally, ```source ~/.profile```. You'll need to re-source every time you use a new shell.
```
CRKatri, line 31: `.profile` will get sourced by every new bash instance iirc, otherwise you could just put it in `.bashrc` or `.zshrc`

LINUX.md
```diff
@@ -0,0 +1,46 @@
+# Building on Linux
+Building on linux is made possible by cctools-port.
+
+Here is how, exactly, to install cctools and use it to compile.
+
+1. Dependencies
+- You need: clang (>=3.4) and everything Procursus normally needs. Get clang with apt.
+
+2. Get iOS SDK
+- You're going to need an iOS 13 SDK. Since you're on Linux, you're going to have to search online for one. I personally use https://github.com/xybp888/iOS-SDKs.
```
CRKatri, line 10: It's best to use an unpatched SDK, you'll also want to copy the c++ stdlib into in, which you will get from the Xcode tool chain

LINUX.md
```diff
@@ -0,0 +1,46 @@
...
+
+3. Compress SDK
+- Compress iPhoneOS.sdk to iPhoneOS.sdk.tar.xz. On Ubuntu you can use the Files program and compress files by right-clicking and selecting Compress.
```
CRKatri, line 13: Is `tar caf iPhoneOS13.5.sdk.xz iPhoneOS13.5.sdk` so hard to run?

LINUX.md
```diff
@@ -0,0 +1,46 @@
...
+- And finally, ```source ~/.profile```. You'll need to re-source every time you use a new shell.
+
+7. Get MacOS SDK
+
+Download the MacOS 10.15 SDK from https://github.com/phracker/MacOSX-SDKs and extract it, then move it to ~/cctools/SDK/MacOSX.sdk.
+
+You can now compile Procursus... with some tweaking. You'll need to set some options manually with your custom build.
+
+You need to set GNU_HOST_TRIPLE=aarch64-apple-darwin11 or whatever cctools generated for you (check the names of files in ~/cctools/bin).
+Also set TARGET_SYSROOT=$(HOME)/cctools/SDK/YOUR_SDK_HERE (in case you didn't use the 13.2 SDK, which is default for this repo).
```
CRKatri, line 40: Just add it to your `.bashrc` or `.zshrc`

CRKatri (changes requested):
Please take a look at [this script by Conan](https://gist.github.com/1Conan/4347fd5f604cfe6116f7acb0237ef155) for the proper way to set up building for Linux.

LINUX.md
Randomblock1: The problem is, it's by default aarch64-apple-darwin, and the files generated by cctools are aarch64-apple-darwin11. That 11 is important. With this, it _does_ build correctly. But if we really shouldn't touch it, I guess we need to rename the cctools then from aarch64-apple-darwin11 to aarch64-apple-darwin.

CRKatri: While somethings may build correctly with that triple, not everything will, some require it to be `aarch64-apple-darwin`, such as openssl

LINUX.md
Randomblock1: I have no idea how or where to get that. Please suggest an edit on exactly how.

CRKatri: Take a look at the script I sent, it has all this taken care of.

LINUX.md
Randomblock1: Fixed it, used sed to rename files properly

LINUX.md
Randomblock1: Fixed

LINUX.md
```diff
@@ -0,0 +1,55 @@
...
+fi
+```
+- And finally, ```source ~/.the_file_you_just_edited```. Or, just exit and relogin to a new shell.
+
+7. Rename files
+
+- You need to rename the files for them to work properly with Procursus.
+
+```cd ~/cctools/bin
+for i in aarch64*; do; mv -v "$i" "$(echo "$i" | sed -e 's/11//' -)"; done
```
CRKatri, line 40: Don't move the files, that will break the symlinks, edit `build.sh`

CRKatri (changes requested):
You're using sed in the wrong place

Randomblock1:
OK, I've remade it and tested it, and it works fine.

CRKatri (approved):

PR #222, opened 2020-11-29
Add harfbuzz and libass

CRKatri:
Untested ~~(as with all my working ones :P)~~ (kinda tested)


Diatrus:
harfuzz fails to build.

Log here: https://hastebin.com/igobubetiz.sh

PR #231, opened 2020-12-05
Add xpwn

marijuanARM:


build_patch/xpwn/install_path.patch
```diff
@@ -0,0 +1,76 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 634cc4f..79fa3d3 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -25,8 +25,8 @@ add_subdirectory (dfu-util)
+ add_subdirectory (xpwn)
+ add_subdirectory (idevice)
+ 
+-install(FILES README.markdown DESTINATION . RENAME README.txt)
+-install(FILES LICENSE DESTINATION . RENAME LICENSE.txt)
```
Diatrus, line 10: One thing I am concerned about with regards to this patch is its portability to other PREFIXes. Can you change this to be more compatible with something like /opt/procursus?

Diatrus (changes requested):

PR #238, opened 2020-12-10
Add img4lib

marijuanARM:


Diatrus:
Package version comes out to:
```
libimg4-dev_1.0+git20201209._iphoneos-arm.deb
```
which, I assume, is incorrect.

PR #240, opened 2020-12-11
Debianize pacman

CRKatri:


Diatrus:
Not needed since the library is only used by pacman.

PR #247, opened 2020-12-15
remove tcg interpreter from qemu's Makefile

asdfugil:
Linux does not boot in tcg interpreter mode plus it is slower.


asdfugil:
Superseded by https://discord.com/channels/715793185019920405/781041045760770049/788437905349017681 (Hayden's Server)

PR #256, opened 2020-12-20
[WIP] Automatic Dependency Generator

CRKatri:
- [ ] Finish writing `build_info/dylibs`
- [ ] Convert *most* control files (optional)
- [ ] Test

Any help on this would be greatly appreciated.


CRKatri:
I figure I should give some information on how to use this. So `build_info/dylibs` will contain all the information on dylibs correspond to what packages, this will be in the format `<target> <dylib> <pkgname>`, so take `berkeleydb`, the target is `berkeleydb` so that's the first entry, then we have the name of the dylib, for example `libdb-18.1.dylib` then we have the package name `libdb18.1`, we do this for each dylib that can be linked so we end up with
```
berkeleydb libdb-18.1.dylib libdb18.1
berkeleydb libdb_cxx-18.1.dylib libdb18.1++
berkeleydb libdb_stl-18.1.dylib libdb18.1-stl
```
When writing your `control` file if you add `@DYLIBS@` to the `Depends:` field all the package names will be added, with a few exceptions, say you're packaging `ffmpeg`, which has multiple libraries, such as `libavcodec58` and `libavdevice58`, these will not be added, so the resulting `Depends` field should look like `Depends: libavcodec58 (= @DEB_FFMPEG_V@), libavdevice58 (= @DEB_FFMPEG_V@), libavfilter7 (= @DEB_FFMPEG_V@), libavformat58 (= @DEB_FFMPEG_V@), libavresample4 (= @DEB_FFMPEG_V@), libpostproc55 (= @DEB_FFMPEG_V@), libswresample3 (= @DEB_FFMPEG_V@), libswscale5 (= @DEB_FFMPEG_V@), libavutil56 (= @DEB_FFMPEG_V@), @DYLIBS@`, this is because when depending on a lib provided by the same package you need the version to be the same, which this will not detect and do automatically. If you also have a package that depends on specific versions of other packages it is best to just write the `Depends` field manually. My current objective is to get `build_info/dylibs` fully made, transitioning the `control` files can we done when updating a package or adding a new package, as the current setup still works, this is more for convenience (and to prevent me from making typos).

PR #279, opened 2020-12-29
Update guide about rust&cargo

suclogger:


Diatrus:
There's a warning in the build_essential package. Maybe it should reference that package instead?

suclogger:
> There's a warning in the build-essential package. Maybe it should reference that package instead?

Right!

PR #284, opened 2020-12-31
add fish

Finermeerkat137:
This is the fish shell from https://fishshell.com


CRKatri:
Has this been tested?

Finermeerkat137:
Yes, I've tested this on my iOS device and it works as expected.

Diatrus:
Thank you for the first contribution, it is much appreciated. 👍

Issue #289, opened 2021-01-03
libpng16-dev should depend on zlib


deatondg:
Running `pkg-config --libs libpng` results in 
```
Package zlib was not found in the pkg-config search path.                                                                                                                                  
Perhaps you should add the directory containing `zlib.pc'                                                                                                                                  
to the PKG_CONFIG_PATH environment variable                                                                                                                                                
Package 'zlib', required by 'libpng', not found          
```
but zlib is not in Procursus. 

I will fix this with a PR (that also fixes a few other small things) in a few minutes.

CRKatri:
zlib is an apple provided lib, we don't have pkgconfig stuff for it, so just edit the libpng.pc to remove the zlib dep and add -lz to you LDFLAGS.

deatondg:
Oh, I see. I already have a copy of zlib building in Procursus. Is it worth submitting that?

CRKatri:
Please do not, it will conflict with apple, you could submit a patch to remove the zlib dependency from libpng.pc however.

deatondg:
Gotcha. Alternatively, would it be a good idea to make a zlib package which just provides the pkgconfig file so that anything else depending on zlib from pkg-config will know where to find it?

CRKatri:
I think it'd be better to just patch the pkgconfig files to remove it as a dep but leave the stuff to link it. Although having a pkgconfig for zlib may not be a bad idea if a configure script is looking for it...

deatondg:
Makes sense to me. Thanks for the info! I appreciate it.

On my system, a configure script is looking for it, but I can just make that for myself until someone else requests it. I’ll make a PR which fixes the pkgconfig file from libpng in a little bit.

PR #290, opened 2021-01-03
Fix libpng.pc, use GNU chown to avoid base-package issues on Big Sur, update build requirements in README.md

deatondg:
This PR includes the following changes:
- Patches libpng.pc.in to use Apple’s zlib instead of looking for zlib.pc (fixes issue #289)
- Bumps the deb version of libpng16
- Uses GNU chown to build base-package on macOS when present
- Update the build-deps in the README


Diatrus:
Why is GNU chown needed on Big Sur?

CRKatri:
Also, python 3.8 is no longer a dependency, we now need python 3.9.

deatondg:
@Diatrus On my system (which is an almost entirely clean Big Sur except for these build dependencies), running `$(FAKEROOT) $(CHOWN) 0:3 $(BUILD_DIST)/base/var/empty` or similar commands results in a permissions error. Switching to GNU’s chown doesn’t have this problem. It’s possible something other than Big Sur causes this, although this worked fine on my old Catalina setup without using GNU chown.

@CRKatri I’ll fix that tomorrow morning.

Diatrus:
@deatondg have you possibly not disabled SIP? fakeroot does not function properly if SIP is enabled on Catalina+. You can test by running `fakeroot whoami`.

deatondg:
@Diatrus Lol... I totally forgot to disable SIP on this install, that’s the problem. Excuse my lack of thought, I’ll revert that commit.

Diatrus:
Thank you tons for your contribution. It's much appreciated, I hope to keep seeing more. 👍

PR #293, opened 2021-01-04
Fix libharfbuzz-dev and libass-dev to refer to libfreetype-dev

deatondg:
instead of libfreetype6-dev.

Currently, it is impossible to install libharfbuzz-dev and libass-dev because they have a dependency on libfreetype6-dev, which does not exist. This package is in Procursus, but it is called libfreetype-dev.

I can bump the deb version if needed.


Diatrus:
Thank you, I cannot believe I missed this.

PR #298, opened 2021-01-10
Convert build_base to tbd instead of dylib in order to save space

CRKatri:
It's not too useful, however I was able to reduce my build_base by 22% by using this. I completely understand if you don't want it though.


CRKatri:
This did not work out, and is causing me a butt load of problems. I believe it is due to a discrepancy in tbd versions, and using unsupported features, but it's not worth the hassle, so I am closing.

PR #302, opened 2021-01-14
Add Neovim and deps

CRKatri:
This will be updated soon to use lua instead of luajit in lua-luv.


CRKatri:
This requires #308 to be merged first.

PR #356, opened 2021-01-22
Update gnupg to version 2.2.27

DaOfficialRolex:


DaOfficialRolex:
The commit text has the wrong version. This is version 2.2.27.

PR #370, opened 2021-01-23
Update libboost to version 1.75.0 and fix libboost dependencies in various control files and in aptitude

DaOfficialRolex:


Diatrus:
I think we should avoid frequent updates to boost. Maybe once every few soversion. We will discuss on the Discord.

PR #385, opened 2021-01-28
New package: bender

aspenluxxxy:
Adds my libhooker configuration CLI, [bender](https://github.com/aspenluxxxy/bender)


Diatrus:
If you wouldn't mind, can you add copying xpc headers from the MACOSX SDK to BUILD_BASE in the main Makefile like is done for many other headers in the setup: stage?

CRKatri:
They are already being copied over, however it seems one of the rust deps isn't respecting the CFLAGS(?).

aspenluxxxy:
> If you wouldn't mind, can you add copying xpc headers from the MACOSX SDK to BUILD_BASE in the main Makefile like is done for many other headers in the setup: stage?

it already did that

CRKatri (approved):

PR #399, opened 2021-02-03
Add libpugixml

absidue:


libpugixml.mk
```diff
@@ -0,0 +1,63 @@
...
+	touch $(BUILD_WORK)/libpugixml/.build_complete
+endif
+
+libpugixml-package: libpugixml-stage
+	# libpugixml.mk Package Structure
+	rm -rf $(BUILD_DIST)/libpugixml{1,-dev}
+	mkdir -p $(BUILD_DIST)/libpugixml{1,-dev}/usr/lib
+
+	# libpugixml.mk Prep libpugixml1
+	cp -a $(BUILD_STAGE)/libpugixml/usr/lib/libpugixml.{1,$(LIBPUGIXML_LIB_VERSION)}.dylib $(BUILD_DIST)/libpugixml1/usr/lib
```
Diatrus, line 46: Instead of a lib version variable, why not do something like `libpugixml.1*.dylib`

Diatrus (changes requested):

PR #412, opened 2021-02-13
Add pw

CRKatri:
- Add pw, adduser, and rmuser to system-cmds
- Fix profile.d and system-cmds postinst

I recommend /etc/shells PR first, but this will still work without it


CRKatri:
FYI this supersedes #411 because of merge conflicts that I was too lazy to fix correctly

PR #423, opened 2021-02-19
Small changes to control files

TheRealKeto:
- Adds a name field to openssh and git.
- Changes the suggested field of gdbmtool, which points to the correct locale package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ ifeq ($(firstword $(subst ., ,$(MAKE_VERSION))),3)
 $(error Install latest make from Homebrew - brew install make)
 endif
 
-ifeq ($(shell /usr/bin/env bash --version | grep -q 'version 5' && echo 1),1)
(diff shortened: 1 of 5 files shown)


CRKatri:
I understand the change of the `Suggests` field, but see no reason for the `Name` field, especially considering that it is not even recognized by Debian.

TheRealKeto:
While it's true that (according to [Debian's control file policy](https://www.debian.org/doc/debian-policy/ch-controlfields.html)) the ``Name`` field doesn't exist, its function targets package managers with a user interface. There's no problem on adding such field, since it has no effect on CLI package managers, meaning that ``openssh`` can still be installed by running ``sudo apt install openssh`` despite the change.

TheRealKeto:
The latest commit I pushed to this PR reverts back the changes I made to the control files of ``openssh`` and ``git``, as per request of Hayden. This can be merged into the project, since all guidelines required are met.

PR #435, opened 2021-02-24
appuninst M1

DaOfficialRolex:


Diatrus:
This is good except make it so it won't build on M1 at all (See: adv-cmds)

PR #436, opened 2021-02-24
Multiplatform DEBIAN files

CRKatri:


Diatrus:
diffutils isn't needed on dpkg at all as far as I can remember. Please confirm by checking the scripts in it, though.

PR #440, opened 2021-02-24
libboost M1

DaOfficialRolex:


DaOfficialRolex:
Major rewrite need to happen for x86_64. Might even have to consider making a separate package for libboost on x86_64. However I might be totally wrong. Cross compilation was successful for arm64 and even the newer version needed no additions to the makefile.

Diatrus (changes requested):
Needs revert to 1.74.0 for now

Diatrus:
Need to fix x86_64 compile

PR #447, opened 2021-02-24
Update nano.mk

usrnamewastaken:
Update GNU nano to 5.6


Diatrus:
We have updated on the `as2` branch, which is where all current development is happening for now.

PR #468, opened 2021-03-01
rbw iphoneos

sunflsks:


sunflsks:
I have a seperate branch based on `as2`, should I merge that as well or will it be autoat

quiprr:
All development is currently doing on the `as2` branch. As long as you only committed these two changed files, you should be able to change the head branch and all will compile fine, but before, change all instances of `usr` in the Makefile to `$(MEMO_PREFIX)/$(MEMO_SUB_PREFIX)` as seen in the most recent pull requests.

sunflsks:
Ok, thank you.

PR #472, opened 2021-03-03
OpenPAM ssh and macOS openssh progress

DaOfficialRolex:


Diatrus:
Needs to be properly fixed for macOS support (using PAM and such)

DaOfficialRolex:
extrainst_ still needs work and also if /var/ is not being used for anything we should remove it to prevent issues with read only directories on macOS. OpenSSH-server looks like the final stretch for this to work properly on macOS systems.

PR #476, opened 2021-03-07
MacOS Darwin-based gpatch

brennanleblanc0:
GNU patch


Diatrus:
Please don’t make a whole new makefile for gpatch. See coreutils.mk or tar.mk on how to do gnutools for macOS.

PR #490, opened 2021-03-09
Update wget to 1.21.1 and added new flag

DaOfficialRolex:
This link should provide context as to why the new flag is needed for the new version wget.

https://www.mail-archive.com/bug-wget@gnu.org/msg09945.html


Diatrus:
@DaOfficialRolex believe it or not, I think I know how to fix that regex error without the flag. If you want to try to fix it yourself, have a look at findutils.

PR #500, opened 2021-03-09
Darwin based Ccache for MacOS

DaOfficialRolex:
CCACHE_CMAKE_ARGS := -DHAVE_NEON=TRUE

This flag is needed for compiling on arm platforms so it is needed for iphoneos-arm64 and darwin-arm64. The build will fail for darwin-amd64 if enabled.


DaOfficialRolex:
At the moment having issues with compiling correctly. By default ccache set SIMD extensions for x86-64 to true while setting ARM's SIMD extension NEON to false. You can correctly compile the program without the flags perfectly fine on x86-64. ARM needs the flags to be set to false while have NEON set to true. Since that is the only SIMD extension that ccache uses and is available on ARM.

DaOfficialRolex:
@1Conan pointed out that ccache set flags by detecting the host system. This mean flags would have to work regardless of what the host system is being used for compiling. The initial method of finding whether that arm64 or amd64 is being called in the command. Should be used to set the flags.

PR #526, opened 2021-03-18
Updated Golang and fixed most of it for MacOS 

DaOfficialRolex:


DaOfficialRolex:
Just a reminder AMD64 compilation hits snag in current state.

PR #530, opened 2021-03-19
Updated sudo and fixed most of it for MacOS

DaOfficialRolex:


DaOfficialRolex:
`sudo.mk:30: *** recipe commences before first target.  Stop. `

Getting absolutely fucked by this stupid error.

PR #533, opened 2021-03-19
Add fd

sunflsks:
The Cargo.toml patch is needed cause upstream for rust-users hasn't added the proper flags to allow building for iOS


sunflsks:
just realised that this had already been added, closing now

PR #534, opened 2021-03-19
add dash shell

sunflsks:
Tested to work on darwin-amd64 and iphoneos-arm64


Diatrus:
I’m going to implement this in order to use it for /bin/sh. I’ll need to do proper tests.

PR #537, opened 2021-03-20
Critical fix for gzip

TheRealKeto:
Since the new updates to ``file-cmds``, users were affected by a small, but pretty significant "restriction", which removed Filza and other dependencies once upgrading. This PR fixes that by removing such "restriction" from ``gzip``'s control file, allowing affected users to install Filza again without issues.


Diatrus:
Thank you

PR #539, opened 2021-03-20
add upx

sunflsks:
Adds UPX and UCL, which is the compression library UPX uses. I can't seem to get UCL to link dynamically while cross compiling for some reason, so i just made it static and added just a -dev package for the lib. I'll try to see if i can get it to dynamically link.


Diatrus:
Thank you for the PR 👍

009647800333145:
Iraq

PR #546, opened 2021-03-22
add jtool2

sunflsks:


CRKatri:
Let me know when this is open source and I'll approve, but I don't think we should distribute binaries that we don't build.

sunflsks:
Yeah, I was thinking that that might be a problem. I'll close this now.

demhademha:
In the words of Hayden regarding this:
`it can i guess`

Diatrus:
Wait till I can test on m1

PR #555, opened 2021-03-23
Update apt to 2.3.1 (add DPKG and APT vendors)

CRKatri:


Diatrus:
Does this also account for the APT vendor? See the cmake args vendor=debian.

PR #560, opened 2021-03-25
Add git-filter-repo

CRKatri:


build_patch/git-filter-repo/autopkgtest--version
```diff
@@ -0,0 +1,32 @@
...
++++ b/t/t9390-filter-repo.sh
+@@ -1670,7 +1670,12 @@ test_expect_success 'tweaking just a tag' '
+ 
+ test_expect_success '--version' '
+ 	git filter-repo --version >actual &&
+-	git hash-object ../../git-filter-repo | colrm 13 >expect &&
++	if [ -e ../../git-filter-repo ]; then
++		filter_repo=../../git-filter-repo
++	else
++		filter_repo=/usr/lib/python3/dist-packages/git_filter_repo.py
```
Diatrus, line 27: This is not prefix friendly. You could change it to an @@ and sed it out during -setup.

Diatrus (changes requested):

Diatrus:
Unsetting MACOSX_VERSION_MIN won't work when compiling for macOS, will it?

git-filter-repo.mk
```diff
@@ -0,0 +1,45 @@
...
+endif
+
+SUBPROJECTS             += git-filter-repo
+GIT_FILTER_REPO_VERSION := 2.29.0
+DEB_GIT_FILTER_REPO_V   ?= $(GIT_FILTER_REPO_VERSION)
+
+git-filter-repo-setup: setup
+	-[ ! -f "$(BUILD_SOURCE)/git-filter-repo-$(GIT_FILTER_REPO_VERSION).tar.gz" ] && \
+		wget -q -nc -O$(BUILD_SOURCE)/git-filter-repo-$(GIT_FILTER_REPO_VERSION).tar.gz \
+			https://github.com/newren/git-filter-repo/releases/download/v$(GIT_FILTER_REPO_VERSION)/git-filter-repo-$(GIT_FILTER_REPO_VERSION).tar.xz
```
TheRealKeto, line 12: Use ``GITHUB_ARCHIVE`` rather than downloading a release from the Releases tab. No additional files are provided by the release, so it's unnecessary to use them.

build_info/git-filter-repo.control
```diff
@@ -0,0 +1,21 @@
+Package: git-filter-repo
+Version: @DEB_GIT_FILTER_REPO_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
```
TheRealKeto, line 4: Add the author of the package below this line.

TheRealKeto (changes requested):
Unsure if this is still being worked on; there are more changes needed to keep up with times...

git-filter-repo.mk
CRKatri: `GITHUB_ARCHIVE` doesn't support this form of release archive.

marruffohomero (approved):

git-filter-repo.mk
TheRealKeto: ``GITHUB_ARCHIVE`` does support tags, and like stated before, the specific release provided by the repository provided no additional files, so we can me "our own" archive. 

If you're so confused about this, the following line is what should replace all that mumbo jumbo; the function exists for you to use it...
```make
$(call GITHUB_ARCHIVE,newren,git-filter-repo,$(GIT_FILTER_REPO_VERSION),v$(GIT_FILTER_REPO_VERSION)
```

CRKatri: The release tarball that I am downloading contains the manpage, which `GITHUB_ARCHIVE`'s tarball will not have.

TheRealKeto: Manpages can still be made with the [Makefile](https://github.com/newren/git-filter-repo/blob/main/Makefile#L26-L40), so this isn't really a reason to not use the function.

While it requires git to do so, most people have git installed on their system to deal with Procursus in general, so this isn't an excuse either...

CRKatri: It requires you to clone the repo, which would take more lines than my wget currently does.

PR #561, opened 2021-03-25
pacman for macOS and converted to meson for future

DaOfficialRolex:


DaOfficialRolex:
bash-completion is also needed for build process, but can build okay without it.

PR #563, opened 2021-03-25
Added bash-completion for those who still use bash

DaOfficialRolex:


brennanleblanc0:
i feel personally attacked by this PR name

demhademha (changes requested):
It should be:
		--prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)
**not**
		--prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX) \

DaOfficialRolex:
> It should be:
> --prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)
> **not**
> --prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX) \

This change doesn't really effect anything. It compiles and packages everything in the same way.

Diatrus:
> > It should be:
> > --prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)
> > **not**
> > --prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX) \
> 
> This change doesn't really effect anything. It compiles and packages everything in the same way.

It definitely does affect. The first make command doesn't run due to it and is passed to configure lol

DaOfficialRolex:
> > > It should be:
> > > --prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)
> > > **not**
> > > --prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX) \
> > 
> > 
> > This change doesn't really effect anything. It compiles and packages everything in the same way.
> 
> It definitely does affect. The first make command doesn't run due to it and is passed to configure lol

Got it will push the change then.

PR #577, opened 2021-03-29
Updating Pacman to 6.0.0 

DaOfficialRolex:


build_patch/pacman/config.diff
```diff
@@ -0,0 +1,44 @@
...
+diff -urN pacman-5.2.2/meson.build pacman/meson.build
+--- pacman-5.2.2/meson.build	2020-06-25 22:04:50.000000000 -0700
++++ pacman/meson.build	2021-03-25 10:47:29.000000000 -0700
+@@ -150,7 +150,7 @@
+ 
+ foreach member : [
+     ['struct stat', 'st_blksize', '''#include <sys/stat.h>'''],
+-    ['struct statvfs', 'f_flag', '''#include <sys/statvfs.h>'''],
++    ['struct statfs', 'f_flags', '''#include <sys/mount.h>'''],
+     ['struct statfs', 'f_flags', '''#include <sys/param.h>
```
Diatrus, line 30: This can surely be done better. Right below here we have a sys/mount.h include.

build_patch/pacman/config.diff
```diff
@@ -0,0 +1,44 @@
...
+                                     #include <sys/mount.h>'''],
+   ]
+@@ -173,8 +173,8 @@
+   endif
+ endforeach
+ 
+-if conf.has('HAVE_STRUCT_STATVFS_F_FLAG')
+-  conf.set('FSSTATSTYPE', 'struct statvfs')
++if conf.has('HAVE_STRUCT_STATVS_F_FLAG')
```
Diatrus, line 39: this is technically invalid

Diatrus (changes requested):
Please learn how to manually set meson build options through the command line without a patch

DaOfficialRolex:
> Please learn how to manually set meson build options through the command line without a patch

Meson doesn't seem to have a conceivable easy way to do this for the options I want to set. Until I can find a way of doing this a patch seems necessary.

DaOfficialRolex:
Need to rediff the patches for this update and then it should work in theory.

makefiles/pacman.mk
```diff
@@ -3,27 +3,44 @@ $(error Use the main Makefile)
 endif
 
 SUBPROJECTS    += pacman
-PACMAN_VERSION := 5.2.2
-DEB_PACMAN_V   ?= $(PACMAN_VERSION)-2
+PACMAN_VERSION := 6.0.0
+DEB_PACMAN_V   ?= $(PACMAN_VERSION)
 
 pacman-setup: setup
 	wget -q -nc -P $(BUILD_SOURCE) https://git.archlinux.org/pacman.git/snapshot/pacman-$(PACMAN_VERSION).tar.gz
```
TheRealKeto, line 10: Get the tarball from [here.](https://sources.archlinux.org/other/pacman/) The signature for the tarball is also there, so add that to the Makefile as well.

makefiles/pacman.mk
```diff
@@ -3,27 +3,44 @@ $(error Use the main Makefile)
...
+	cd $(BUILD_WORK)/pacman/build && PKG_CONFIG="pkg-config" meson \
+		--cross-file cross.txt \
+		--buildtype=release \
+		-D i18n=false \
+		-D doc=disabled \
+		-D crypto=openssl \
+		-D b_bitcode=true \
+		..
+	+ninja -C $(BUILD_WORK)/pacman/build
+	+DESTDIR="$(BUILD_STAGE)/pacman" ninja -C $(BUILD_WORK)/pacman/build install
```
TheRealKeto, line 43: Change line 43 to the following
```suggestion
	+ninja -C $(BUILD_WORK)/pacman/build install \
		DESTDIR="$(BUILD_STAGE)/pacman"
```

build_info/pacman.control
```diff
@@ -3,7 +3,7 @@ Version: @DEB_PACMAN_V@
 Architecture: @DEB_ARCH@
 Maintainer: @DEB_MAINTAINER@
 Section: Utilities
-Depends: libarchive13, libarchive-tools, openssl, libcurl4, libzstd1, tar, libintl8
+Depends: libarchive13, libarchive-tools, openssl, libcurl4, libzstd1, tar, libintl8, bash-completion
 Priority: optional
 Description: A library-based package manager with dependency support
 Name: makepkg
```
TheRealKeto, line 9: Why is this here to begin with?

TheRealKeto (changes requested):

DaOfficialRolex:
This builds and works correctly. Please review before merging

DaOfficialRolex:
Need to add libgpgme11 to the control file.

DaOfficialRolex:
Need to look at makepkg.

build_info/pacman.control
CRKatri: This is because Hayden didn’t want people to install it and mess up their device, labeling it as `makepkg` well make less people recognize it as the package manager thinking they can use it in parity with `dpkg`, but instead as `makepkg`, which is the only reason to install this.

CRKatri: https://github.com/ProcursusTeam/Procursus/pull/102#issuecomment-673544226

TheRealKeto: This could have been made clear in the description. The name of the package is pacman, and considering that we often follow the naming based on what Debian/Arch uses, the name flag isn't needed.

DaOfficialRolex:
Need to find out why meson hard codes the prefix of bash-completion that is installed on the host system. See if it can be changed to the prefix of the system that we are building for.

DaOfficialRolex:
Fixed my earlier mistake. Hopefully the last time I ask this to be reviewed

PR #590, opened 2021-04-01
Update youtube-dl to latest version

TheRealKeto:
This PR updates ``youtube-dl`` to the latest released version. This should fix any errors users experience with older installations. Go outside with your friends and prank your boss today!

### DIFF
--- a/youtube-dl.mk
+++ b/youtube-dl.mk
@@ -3,7 +3,7 @@ $(error Use the main Makefile)
 endif
 
 SUBPROJECTS        += youtube-dl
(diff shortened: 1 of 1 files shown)


TheRealKeto:
I've converted this to a draft, since I can't get the patch to correctly apply. I will provide further information at a later time.

TheRealKeto:
**Update**: After some testing, the patch applied correctly and everything seems to work. This is only the case for iOS, as I cannot test for other platforms. It should still be fine to merge,

PR #520, opened 2021-03-16
Add binutils

CRKatri:

PR #591, opened 2021-04-01
Add all of x11 fonts to Procursus 

Gymxo:


CRKatri:
This is the exact opposite of debianization.

TheRealKeto:
This is better than having to make each package separately, which would only add more "weight" by making about 32 Makefiles to the bootstrap, which seems inefficient.

marijuanARM:
The way to go with this would be to have it be one makefile, but separate packages for each font.

fonts.mk
```diff
@@ -0,0 +1,521 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS    += Fonts
```
marijuanARM, line 5: Should be `x11-fonts` instead of `Fonts` (and `x11-fonts.mk` instead of `fonts.mk`)

fonts.mk
```diff
@@ -0,0 +1,521 @@
...
+
+fonts-package: fonts-stage
+# fonts.mk Package Structure
+	rm -rf $(BUILD_DIST)/fonts
+	
+# fonts.mk Prep fonts
+	cp -a $(BUILD_STAGE)/fonts $(BUILD_DIST)
+	
+# fonts.mk Sign
+	$(call SIGN,fonts,general.xml)
```
marijuanARM, line 513: Signing isn't necessary as they're neither executables nor libraries.

fonts.mk
```diff
@@ -0,0 +1,521 @@
...
+	
+# fonts.mk Make .debs
+	$(call PACK,fonts,DEB_FONTS_V)
```
marijuanARM, line 516: The fonts should be split up into separate packages.

fonts.mk
```diff
@@ -0,0 +1,521 @@
...
+		DESTDIR=$(BUILD_BASE)
+	touch $(BUILD_WORK)/font-xfree86-type1/.build_complete
+endif
+
+fonts-package: fonts-stage
+# fonts.mk Package Structure
+	rm -rf $(BUILD_DIST)/fonts
+	
+# fonts.mk Prep fonts
+	cp -a $(BUILD_STAGE)/fonts $(BUILD_DIST)
```
marijuanARM, line 510: The fonts should be split up into separate packages.

build_info/fonts.control
```diff
@@ -0,0 +1,8 @@
+Package: X11-fonts
```
marijuanARM, line 1: Lowercase X for the package name, please.

marijuanARM (changes requested):

fonts.mk
```diff
@@ -0,0 +1,521 @@
...
+	wget -q -nc -P $(BUILD_SOURCE) https://www.x.org/archive/individual/font/font-sun-misc-$(font-sun-misc).tar.gz
+	$(call EXTRACT_TAR,font-sun-misc-$(font-sun-misc).tar.gz,font-sun-misc-$(font-sun-misc),font-sun-misc)
+	wget -q -nc -P $(BUILD_SOURCE) https://www.x.org/archive/individual/font/font-winitzki-cyrillic-$(font-winitzki-cyrillic).tar.gz
+	$(call EXTRACT_TAR,font-winitzki-cyrillic-$(font-winitzki-cyrillic).tar.gz,font-winitzki-cyrillic-$(font-winitzki-cyrillic),font-winitzki-cyrillic)
+	wget -q -nc -P $(BUILD_SOURCE) https://www.x.org/archive/individual/font/font-xfree86-type1-$(font-xfree86-type1).tar.gz
+	$(call EXTRACT_TAR,font-xfree86-type1-$(font-xfree86-type1).tar.gz,font-xfree86-type1-$(font-xfree86-type1),font-xfree86-type1)
+
+ifneq ($(wildcard $(BUILD_WORK)/font-xfree86-type1/.build_complete),)
+fonts:
+	@echo "Using previously built all fonts."
```
TheRealKeto, line 118: Specifically state ``x11-fonts``.

marijuanARM:
@CRKatri anything else you see here that needs to be changed that i missed?

fonts.mk
```diff
@@ -0,0 +1,521 @@
...
+# fonts.mk Sign
+	$(call SIGN,fonts,general.xml)
+	
+# fonts.mk Make .debs
+	$(call PACK,fonts,DEB_FONTS_V)
+	
+# fonts.mk Build cleanup
+	rm -rf $(BUILD_DIST)/fonts
+
+.PHONY: fonts fonts-package
```
TheRealKeto, line 521: Consider adding an empty line at the end of your file.

TheRealKeto (changes requested):
I've nitpicked a few things about your PR, but, in general, this should be fine. References to ``fonts`` (in any regard) should be renamed to ``x11-fonts``.

I also feel like specific parts of this PR can be improved, specifically where and how each package is "retrieved". (Perhaps find/host a mirror of all of these files, that way, you don't need to constantly extract each individual font from "different" source?)

fonts.mk
Diatrus: I don’t have a problem in this case.

Diatrus: Keep it as one.

CRKatri: I think this is a case where splitting is necessarily, I don’t need 32 fonts unless I’m using libreoffice, so all these fonts would be bloat, besides, splitting it would be super easy to do, you wouldn’t even need to create 32 make files, or have repeat lines if you do it the same was as #520

Diatrus: That’s a good point. Agreed.

sunflsks:
> I've nitpicked a few things about your PR, but, in general, this should be fine. References to `fonts` (in any regard) should be renamed to `x11-fonts`.
> 
> I also feel like specific parts of this PR can be improved, specifically where and how each package is "retrieved". (Perhaps find/host a mirror of all of these files, that way, you don't need to constantly extract each individual font from "different" source?)

I think getting the fonts directly from upstream, rather than relying on a third-party mirror, is a better idea, due to issues like trust and availability. However, what could be done is to put the redundant part of the URL (https://www.x.org/archive/individual/font) in a variable and use that instead to help improve readability.

For example,
 	`wget -q -nc -P $(BUILD_SOURCE) $(X_FONT_URL)/font-adobe-utopia-100dpi-$(font-adobe-utopia-100dpi).tar.gz`

Gymxo:
I’ll change some stuff before I pr again

PR #598, opened 2021-04-04
add help2man

demhademha:
Programs that support this should probably be recompiled


CRKatri:
Well this just needs to be installed on the host, so as long as @Diatrus had help2man installed on his computer while building the debs for the repo, no recompilation is needed.

demhademha:
True

PR #600, opened 2021-04-04
automatic dependency generator 2

CRKatri:
Almost identical to #256, except build_misc/dylibs gets generated automatically when you run `tool-package`.


Diatrus:
I would personally hold on this until the rewrite/restructure. Lots of stuff will change.

CRKatri:
Sounds good, I just wanted to get it off of local.

CRKatri:
The ideal way to do this is to add Mach-O support to `dpkg-dev`, but I don’t know Perl

PR #474, opened 2021-03-05
Add theos metapackages

TheRealKeto:
This adds packages required to setup Theos with Procursus.

PR #605, opened 2021-04-05
Add theos metapackages

TheRealKeto:
This PR adds ``theos-dependencies`` and ``ios-toolchain`` to Procursus, allowing users to setup Theos with the right packages. This should also prevent users from installing the BigBoss counterparts of these packages, since those can cause conflicts with other Procursus packages (such as ``clang``).

This is a duplicate of [#474](https://github.com/ProcursusTeam/Procursus/pull/474), but it actually compiles...

### DIFF
--- a/build_info/ios-toolchain.control
+++ b/build_info/ios-toolchain.control
@@ -1,0 +1,14 @@
+Package: ios-toolchain
+Version: @DEB_THEOSDEPS_V@
+Architecture: @DEB_ARCH@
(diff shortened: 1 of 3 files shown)


TheRealKeto:
Closing this since this isn't necessary anymore; theos (and all of their provided packages needed to work) should be fully capable of handling what this PR attempted to do, and handling it "ourselves" isn't the best approach; Hayden privately mentioned that this wouldn't have worked anyways...

PR #606, opened 2021-04-05
add tcsh and rc

sunflsks:
I've also done ksh, but i've had to do some horrendous patching of the `meson.build`, so I'm not going to include that here. I have it on a seperate branch on my fork. rc is fully functional on my iOS 14 device. tcsh is unable to access the termcap database, so that needs to be fixed as well.

### DIFF
--- a/build_info/rc.control
+++ b/build_info/rc.control
@@ -2,7 +2,7 @@ Package: rc
 Version: @DEB_RC_V@
 Architecture: @DEB_ARCH@
 Maintainer: @DEB_MAINTAINER@
-Depends: libreadline8
+Depends: libreadline8, libncursesw6
 Section: Terminal_Support
 Priority: optional
 Homepage: http://tobold.org/article/rc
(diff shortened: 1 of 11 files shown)


sunflsks:
The problem with `tcsh` not being able to find the terminfo database does not seem to be in the way I've compiled it, but rather due to the fact that there is something wrong with the terminfo database, since `tcsh` from other repositories (notably MCApollo) seem to have the same issue. Therefore, I'm marking this as ready to review.

Diatrus:
@sunflsks does it successfully link with libncursesw from Procursus? or does it link the one from the macOS sdk?

sunflsks:
I just ran `otool -L` on the tcsh binary, and it turns out it's linking to `libncurses.5.4.dylib`, which is probably the ncurses from my macOS machine.

sunflsks:
This commit fixes the issue with tcsh so it should be ready to merge

PR #610, opened 2021-04-07
allow executing shell scripts in fish

sunflsks:
This allows for executing shell scripts with a shebang on iOS. The code quality on this is not very good, but it's meant as more of a stopgap measure until a later release of fish, which will hopefully add a more robust and elegant way to do this (it's already in upstream master)


sunflsks:
just my luck, fish 3.2.2 released 5 hours ago. I'll close this now and use the better method of doing this

sunflsks:
nevermind, turned out it hasn't been merged in the latest release yet

PR #611, opened 2021-04-08
add haydenfetch

asdfugil:
fr
Tested buidling and running on iPadOS 14.3, iPad Pro 9.7 Inch.

### DIFF
--- a/build_info/haydenfetch.control
+++ b/build_info/haydenfetch.control
@@ -1,7 +1,8 @@
 Package: haydenfetch
 Architecture: @DEB_ARCH@
 Version: @DEB_HAYDENFETCH_V@
-Depends: jp2a, coreutils, neofetch
+Pre-Depends: coreutils
+Depends: jp2a, neofetch
 Suggests: imagemagick
 Description: neofetch but with haydens
 Maintainer: @DEB_MAINTAINER@
--- a/build_info/haydenfetch.control
+++ b/build_info/haydenfetch.control
@@ -1,0 +1,8 @@
+Package: haydenfetch
+Architecture: @DEB_ARCH@
+Version: @DEB_HAYDENFETCH_V@
+Depends: jp2a, coreutils, neofetch
+Suggests: imagemagick
+Description: neofetch but with haydens
+Maintainer: @DEB_MAINTAINER@
+Author: Nick Chan <towinchenmi@gmail.com>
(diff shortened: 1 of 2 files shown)


build_info/haydenfetch.control
```diff
@@ -0,0 +1,8 @@
+Package: haydenfetch
+Architecture: @DEB_ARCH@
+Version: @DEB_HAYDENFETCH_V@
+Depends: jp2a, coreutils, neofetch
```
TheRealKeto, line 4: ``coreutils`` should instead be in the ``Pre-Depends`` field, since it's a core package

haydenfetch.mk
```diff
@@ -0,0 +1,42 @@
...
+DEB_HAYDENFETCH_V   ?= $(HAYDENFETCH_VERSION)
+
+haydenfetch-setup: setup
+	if [ ! -d "$(BUILD_WORK)/haydenfetch" ]; then \
+	git clone https://github.com/asdfugil/haydenfetch.git  $(BUILD_WORK)/haydenfetch; \
+	cd $(BUILD_WORK)/haydenfetch; \
+	git fetch origin; \
+	git reset --hard origin/master; \
+	git checkout "3d16901a8d0e8699b7d1484f135d7c2aad402ade"; \
+	fi
```
TheRealKeto, line 16: I feel like, because you own this repository, you should make a tag and use that implementation of downloading a tar rather than using this, which depends on ``git``, which is not a Procursus dependency.

haydenfetch.mk
```diff
@@ -0,0 +1,42 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS      += haydenfetch
```
TheRealKeto, line 5: Can you fix the spacing of this for readability?

TheRealKeto (commented):
This is cool! But I left a few comments on this that should be changed before this PR is accepted.

haydenfetch.mk
```diff
@@ -0,0 +1,42 @@
...
+	git checkout "3d16901a8d0e8699b7d1484f135d7c2aad402ade"; \
+	fi
+
+ifneq ($(wildcard $(BUILD_WORK)/haydenfetch/.build_complete),)
+haydenfetch:
+	@echo "Using previously built haydenfetch."
+else
+haydenfetch: haydenfetch-setup
+	+$(MAKE) -C $(BUILD_WORK)/haydenfetch install \
+		PREFIX=/usr \
```
brennanleblanc0, line 24: I feel this PR should probably be directed to AS2 branch

In that case, the prefix would need to use MEMO_PREFIX and MEMO_SUB_PREFIX

brennanleblanc0 (changes requested):

asdfugil:
Superseded by #612 .

PR #616, opened 2021-04-09
Default configure/cmake flags

1Conan:


CRKatri:
Add go flags too, especially since Darwin and iOS have different GOOS now.

Issue #229, opened 2020-12-01
Unable to build various complex project on-device due to various `Operation not permitted` errors.


deatondg:
I am not savvy enough to understand exactly what is going wrong here, so please let me know if I should submit this issue to the Odyssey repo or some other repo instead. 

I am unable to build various complex projects on-device (iPad Pro 4 iOS 13.5 Odyssey 1.2.2) due to various `Operation not permitted` errors. I had the same issue on Unc0ver using Sam Binger’s tools, and it was acknowledged as a bug [on StackOverflow](https://stackoverflow.com/questions/62359060/why-do-builds-for-various-projects-fail-with-operation-not-permitted-using-ios), but never fixed. I’ve recently switched to Odyssey and your tools.

The errors are all something like this: building `make` using
```
git clone https://git.savannah.gnu.org/git/make.git
cd make
./bootstrap
```
fails with
```
autoreconf: running: aclocal -I m4 --force -I m4
Can't exec "aclocal": Operation not permitted at /usr/share/autoconf/Autom4te/FileUtils.pm line 326.
autoreconf: failed to run aclocal: Operation not permitted
./bootstrap: autoreconf failed
```

Building texlive using
```
git clone https://github.com/TeX-Live/texlive-source
cd texlive-source
```
fails with
```
/home/mobile/Developer/sources/texlive-source/libs/freetype2/freetype-src/builds/unix/configure  '--disable-shared' '--without-bzip2' '--without-zlib' '--without-png' '--without-harfbuzz' '--prefix=/home/mobile/Developer/sources/texlive-source/Work/libs/freetype2/ft-install' '--libdir=/home/mobile/Developer/sources/texlive-source/Work/libs/freetype2' '--includedir=/home/mobile/Developer/sources/texlive-source/Work/libs/freetype2'
make[4]: /home/mobile/Developer/sources/texlive-source/libs/freetype2/freetype-src/builds/unix/configure: Operation not permitted
make[4]: *** [/home/mobile/Developer/sources/texlive-source/libs/freetype2/freetype-src/builds/unix/detect.mk:89: setup] Error 127make[4]: Leaving directory '/home/mobile/Developer/sources/texlive-source/Work/libs/freetype2/ft-build'make[3]: *** [Makefile:606: ft-config] Error 2
make[3]: Leaving directory '/home/mobile/Developer/sources/texlive-source/Work/libs/freetype2'make[2]: *** [Makefile:918: recurse] Error 1
[...]

Issue #226, opened 2020-12-01
tmux appears broken


deatondg:
I cannot get tmux to work. Running `tmux` on my iPad Pro 4 results in
```
[server exited unexpectedly]
                            %                                                                                                                                                              Daviss-iPad:~ mobile% 
/1;2c_
```
I think this also used to work, but I don’t remember when. Let me know if there’s anything I can do to help debug.

Sidenote: thanks for fixing my last issue in less than 10 min. You’re incredible.

Diatrus:
Fixed in 0e51f487c408e9cc7f25a81c15420ca1e5533bb8

To the sidenote: no problem at all.

PR #619, opened 2021-04-10
GNU Make EPERM patch

sunflsks:
I guess make sometimes executes commands directly instead of in a sub shell, so it needs an EPERM patch as well


CRKatri:
Could this be related to #226?

sunflsks:
Yes, this and #229 address the same problem.

PR #622, opened 2021-04-10
Xorg-server

Gymxo:


CRKatri:
Is this use able? or still WIP.

Gymxo:
> Is this use able or still WIP?

Yes this works

CRKatri:
Over VNC, or can it drive the display?

asdfugil:
afaik it can't use the framebuffer (yet)

Gymxo:
The TigerVnc to which currently isn't accepted yet has Xvnc to which can create virtual displays that can be viewed over vnc viewers. However, Xnest and Xephyr still work when doing export DISPLAY and can be used to display apps. The reason why Xorg and Xdmx got added is that I was told by Hayden to add them(even if they don't work properly atm).

PR #624, opened 2021-04-11
Add GNU hello

asdfugil:
I am surprised that this isn't in Procursus yet.


asdfugil:
converting to draft because I want to patch it to say `Hello, Procursus!`.

PR #627, opened 2021-04-11
Fixes tihmstar project configure versioning

### LINKED ISSUE #626: xpwn base is the wrong path
Because the path is wrong, the libs and headers are being stored nested in one more xpwn dir.

Cryptiiiic:
Fixes tihmstar project configure versioning


Diatrus:
Don’t change the xpwn includes path.

Cryptiiiic:
> Don’t change the xpwn includes path.

does 69c494b look good for this? @Diatrus

Diatrus:
Better. Putting on hold for discussion, as @1Conan had another idea for doing this.

Cryptiiiic:
I have local changes for versioning that I will push soon, its all sed patches for tihmstar configure projects @Diatrus

Cryptiiiic:
I've done all I can think of for the time being, unless there is more for xpwn.

Cryptiiiic:
Undid xpwn, removed comments

PR #643, opened 2021-04-15
Update Golang to 1.16.3

Randomblock1:
The main change in 1.16 is how it renamed the build target for iOS from `darwin/arm64` to `ios/arm64`, because now there's ARM64 Macs and that gets confusing. Of course, there are some bugfixes too.


Randomblock1:
wait nevermind someone already made a pr for this

PR #644, opened 2021-04-15
add carthage

sunflsks:


carthage.mk
```diff
@@ -0,0 +1,44 @@
...
+endif
+
+ifneq (,$(findstring darwin,$(MEMO_TARGET)))
+
+SUBPROJECTS      += carthage
+CARTHAGE_VERSION   := 0.37.0
+DEB_CARTHAGE_V     ?= $(CARTHAGE_VERSION)
+
+carthage-setup: setup
+	wget -q -O $(BUILD_SOURCE)/Carthage-$(CARTHAGE_VERSION).tar.gz https://github.com/Carthage/Carthage/archive/$(CARTHAGE_VERSION).tar.gz
```
Cryptiiiic, line 12: You can use the new as2 branch GITHUB_ARCHIVE Format instead of wget directly for github.

Cryptiiiic (changes requested):

carthage.mk
sunflsks: This has been resolved in the latest commit.

carthage.mk
```diff
@@ -0,0 +1,45 @@
...
```
absidue, line 12: Please remove this line, the download is handled by `GITHUB_ARCHIVE`

absidue (changes requested):

carthage.mk
sunflsks: My bad, I thought I had deleted that line. Just pushed a commit that fixes this

absidue (approved):
Looks good now

PR #659, opened 2021-04-19
Add nethack

Randomblock1:
https://github.com/nethack/nethack

Also, Nethack 3.7 has a whole cross-compile build system thing so it might be a better idea to use that.

### DIFF
--- a/nethack.mk
+++ b/nethack.mk
@@ -3,20 +3,25 @@ $(error Use the main Makefile)
 endif
 
 SUBPROJECTS     += nethack
-NETHACK_VERSION := 3.7
+NETHACK_COMMIT  := 2d2c584c961d4328044f587b1135eef216146d45
+NETHACK_VERSION := 3.7+git20210419.$(shell echo $(NETHACK_COMMIT) | cut -c -7)
 DEB_NETHACK_V   ?= $(NETHACK_VERSION)
 
 nethack-setup: setup
 	rm -Rf $(BUILD_WORK)/nethack
-	git clone --depth 1 --single-branch --branch NetHack-$(NETHACK_VERSION) https://github.com/nethack/nethack $(BUILD_WORK)/nethack
+	$(call GITHUB_ARCHIVE,NetHack,NetHack,$(NETHACK_COMMIT),$(NETHACK_COMMIT))
+	$(call EXTRACT_TAR,NetHack-$(NETHACK_COMMIT).tar.gz,NetHack-$(NETHACK_COMMIT),nethack)
 	mkdir -p $(BUILD_STAGE)/nethack/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin
+	sed -i 's/WINTTYLIB=-lncurses -ltinfo/WINTTYLIB=-lncurses/' $(BUILD_WORK)/nethack/sys/unix/hints/linux
+	sed -i 's,PREFIX=.*/nh/install,'"PREFIX=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)"',' $(BUILD_WORK)/nethack/sys/unix/hints/linux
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -197,7 +197,7 @@ MACOSX_SYSROOT  ?= $(shell xcrun --show-sdk-path)
 CC              := cc
 CXX             := c++
 CPP             := cc -E
-PATH            := /opt/procursus/bin:/opt/procursus/libexec/gnubin:/usr/bin:$(PATH)
+PATH            := /opt/procursus/bin:/opt/procursus/libexec/gnubin:$(PATH):/usr/bin
```
Cryptiiiic, line 200: I don't understand why this is needed. Also try not to patch the main Makefile in a subproject pull.

Cryptiiiic (changes requested):

nethack.mk
```diff
@@ -0,0 +1,48 @@
...
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS     += nethack
+NETHACK_VERSION := 3.7
+DEB_NETHACK_V   ?= $(NETHACK_VERSION)
+
+nethack-setup: setup
+	rm -Rf $(BUILD_WORK)/nethack
+	git clone --depth 1 --single-branch --branch NetHack-$(NETHACK_VERSION) https://github.com/nethack/nethack $(BUILD_WORK)/nethack
```
Cryptiiiic, line 11: Use GITHUB_ARCHIVE with commit sha instead of cloning

Cryptiiiic (changes requested):

Makefile
Diatrus: Agreed.

nethack.mk
```diff
@@ -0,0 +1,48 @@
...
+	mkdir -p $(BUILD_STAGE)/nethack/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin
+
+ifneq ($(wildcard $(BUILD_WORK)/nethack/.build_complete),)
+nethack:
+	@echo "Using previously built nethack."
+else
+nethack: nethack-setup
```
Diatrus, line 18: Require “ncurses” here.

Diatrus (changes requested):

CRKatri:
Install game binaries to `$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/games` and the games scoreboard, etc in `$(MEMO_PREFIX)/var/games/gamename/`

Makefile
Randomblock1: Sorry, I should have made a separate PR.
The reason for this is that when you don't have the bootstrap on a Mac but have tools installed from Brew, it keeps telling you to install GNU tar/dpkg/etc, even if you have them installed and in your PATH.
This is because /usr/bin is prioritized over your PATH, which it shouldn't be, because then it tries to use /usr/bin/make instead of /opt/homebrew/.../make, leading to an error when it could just use the tools installed from Brew to compile just fine.

Diatrus: @Randomblock1 this is intentional. The as2 branch requires tools from Procursus. By the time it gets merged to master, there will be an easy install script + dependencies from Procursus Mac. This is for, in the future, it will be easier to do build dependencies per-project. If you'd care to join the discord, I can probably explain better.

Randomblock1: OK, I reverted the commit. What's the Discord invite?

CRKatri:
Oh, and use `$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/games/gamename/` for other stuff.

Makefile
Diatrus: @Randomblock1 https://diatr.us/discord/

Cryptiiiic:
instead of copying everything from work to stage, shouldn't you just set destdir to stage?

TheRealKeto:
Please rebase this with the latest changes to upstream.

PR #665, opened 2021-04-20
xlsclients for x11

Gymxo:


Gymxo:
This package is included in https://packages.debian.org/sid/x11-utils

PR #670, opened 2021-04-20
Globalize GITHUB_ARCHIVE

TheRealKeto:
This PR transitions all packages coming from Github to the ``GITHUB_ARCHIVE`` function (introduced by absidue in [#419](https://github.com/ProcursusTeam/Procursus/pull/419)). The only exceptions are packages that don't come from Github, or get their tarballs from the projects' Github Release tab. 

From this PR on, new additions should, and (likely) will be required to use the function if the package comes from Github.

### DIFF
--- a/axel.mk
+++ b/axel.mk
@@ -2,7 +2,7 @@ ifneq ($(PROCURSUS),1)
 $(error Use the main Makefile)
 endif
 
-SUBPROJECTS      += axel
+SUBPROJECTS  += axel
 AXEL_VERSION := 2.17.10
 DEB_AXEL_V   ?= $(AXEL_VERSION)
 
--- a/bash-completion.mk
+++ b/bash-completion.mk
@@ -39,4 +39,3 @@ bash-completion-package: bash-completion-stage
 	rm -rf $(BUILD_DIST)/bash-completion
 
 .PHONY: bash-completion bash-completion-package
-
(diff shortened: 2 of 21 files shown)


TheRealKeto:
The packages I've transitioned to the function fully work, except for ``cowsay`` and ``libpam-google-authenticator`` (files don't copy to the correct path). I will push a commit that fixes these packages later.

CRKatri:
You also missed msgpack.

TheRealKeto:
``msgpack`` uses a tarball that comes straight from the project's Release tab. Those released tarballs likely have other files which would otherwise be missing if we created our own tarball.

CRKatri:
I don't understand, does `GITHUB_ARCHIVE` not support release? Cause that should be fixed.

absidue:
@CRKatri `GITHUB_ARCHIVE` doesn't support releases as those tarballs have the correct naming format `repo-version.extension` or `package-version.extension` in most cases, the ones that don't can be handled differently on an individual basis.

The `GITHUB_ARCHIVE` function works around wget naming the tarballs incorrectly when downloding an archive of a branch (`branch.tar.gz`), commit (`commit.tar.gz`) or tag (`tag.tar.gz`) from GitHub by forcing a specific filename (`repo-version.extension` or `$(5)-version.extension` if parameter 5 is specified) and preventing the download if it is already downloaded.

Other tarball sources have similar issues but are much less common in Procursus, so a dedicated function is not really necessary for those at this point in time.

PR #675, opened 2021-04-21
add checksum verify function

sunflsks:
This adds a `CHECKSUM_VERIFY` function. The arguments are as follows:

$1 must contain  the algorithm (sha1, sha256, sha512)
$2 must contain the file to hash
$3 must contain the checksum of the file to be hashed. 

Here is an example:
`$(call CHECKSUM_VERIFY,sha512,$(BUILD_SOURCE)/Tarball.tar.gz,<checksum here>)`


CRKatri:
This is not portable.

sunflsks:
May I ask why? It's 100% POSIX-compliant, and the only assumption it makes is that the host system has `shasum`, which is preinstalled on almost every UNIX.

sunflsks:
I've changed the function to  use `openssl`, which should be preinstalled in most modern Unix systems

CRKatri:
Sadly POSIX doesn't specify a program for checksums, so the best way do it is to look for `sha1` if that doesn't exist, `sha1sum` and finally `gsha1sum`. Then replace sha1 with whatever the hash is (sha256, md5, etc)

sunflsks:
Even though POSIX doesn't specify it, I think it's safe to assume that OpenSSL is installed on basically every modern operating system besides Windows 10 (which I don't think we'll be targeting as a build platform anytime soon, anyway)

Diatrus:
@sunflsks openssl *binaries* currently aren't a depend for building Procursus on Debian or iOS. Could we use the shasum bins?

PR #678, opened 2021-04-21
add libsystem-manpages

sunflsks:
This adds man2 (syscalls), man3 (libc functions), and man4 (device files) manpages


Diatrus:
Please prevent this from building on darwin* as well as use MEMO_PREFIX and such

PR #682, opened 2021-04-23
Add pueue

TheRealKeto:
This PR adds ``pueue``, a tool that allows users to manage their shell commands.

### DIFF
--- a/pueue.mk
+++ b/pueue.mk
@@ -3,7 +3,7 @@ $(error Use the main Makefile)
 endif
 
 SUBPROJECTS   += pueue
-PUEUE_VERSION := 0.12.2
+PUEUE_VERSION := 1.0.0-rc.1
 DEB_PUEUE_V   ?= $(PUEUE_VERSION)
 
 pueue-setup: setup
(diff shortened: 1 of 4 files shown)


TheRealKeto:
Special thanks to [althiometer](https://github.com/althiometer) for adding iOS support to ``pueue``  and the last commit :)

pueue.mk
```diff
@@ -0,0 +1,53 @@
...
+	$(GINSTALL) -Dm755 $(BUILD_WORK)/pueue/target/$(RUST_TARGET)/release/pueue \
+		$(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/pueue
+	$(GINSTALL) -Dm755 $(BUILD_WORK)/pueue/target/$(RUST_TARGET)/release/pueued \
+		$(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/pueued
+	# Setup necessary daemon requirements for pueued
+	mkdir -p $(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/Library/LaunchDaemons
+	cp -a $(BUILD_MISC)/pueue/com.nukesor.pueued.plist \
+		$(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/Library/LaunchDaemons
+	# "Patch" up LaunchDaemon plist for pueued
+	for file in $(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/Library/LaunchDaemons/*; do \
```
m1stadev, line 32: This is unnecessary, the PACK function does this automatically.

pueue.mk
```diff
@@ -0,0 +1,53 @@
...
+else
+pueue: pueue-setup
+	cd $(BUILD_WORK)/pueue && SDKROOT="$(TARGET_SYSROOT)" \
+	PKG_CONFIG="$(RUST_TARGET)-pkg-config" cargo build \
+		--release \
+		--target=$(RUST_TARGET)
+	# Move pueue and pueued binaries to designated places
+	$(GINSTALL) -Dm755 $(BUILD_WORK)/pueue/target/$(RUST_TARGET)/release/pueue \
+		$(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/pueue
+	$(GINSTALL) -Dm755 $(BUILD_WORK)/pueue/target/$(RUST_TARGET)/release/pueued \
```
m1stadev, line 25: These 2 install commands can be combined into one by using `pueue{,d}`.

m1stadev (changes requested):

pueue.mk
```diff
@@ -0,0 +1,53 @@
...
+		$(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/pueued
+	# Setup necessary daemon requirements for pueued
+	mkdir -p $(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/Library/LaunchDaemons
+	cp -a $(BUILD_MISC)/pueue/com.nukesor.pueued.plist \
+		$(BUILD_STAGE)/pueue/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/Library/LaunchDaemons
```
m1stadev, line 30: You shouldn't be using MEMO_SUB_PREFIX here.

m1stadev (changes requested):

TheRealKeto:
I think the best way to make this work (more specifically, to make the daemon start for all users) is to create a user specifically designed for pueued. The daemon, in its current state, only works under the ``mobile`` users, but I feel as if adding a user specifically designed for pueue (much how Gitea was approached) would solve the issue of pueued running in the background only for ``mobile``. I'll see what I can do about it at a later date.

TheRealKeto:
Unsure how this was unmarked as a draft; I'm "still" working on this — just need to find a better solution.

PR #686, opened 2021-04-26
add bottom

althiometer:
add bottom, a rust version of bottomify.


m1stadev:
tested on darwin-amd64 + iphoneos-arm64 1700, lgtm!

PR #690, opened 2021-04-28
Add opendircolors

CRKatri:
This is untested because Procursus doesn't support FreeBSD, and is only on `darwin-*` because it is useless on iOS because you only have a GNU userland, unlike MacOS which comes with a FreeBSD userland.


Diatrus:
I want to replace the GNU userland on iOS properly sometime so we’ll keep this held right now.

CRKatri:
Should now support iOS, although usefulness is pretty much zero, until a BSD userland is available.

demhademha:
I think the section should be **utilities**

CRKatri:
When you say you want a BSD userland do you want to use macOS' from opensource.apple.com? Or porting it from (Free|Net|Open)BSD which I don't like because apple did most the work and has some of it's own features.

PR #692, opened 2021-04-30
Add apt and dpkg dependencies to the bootstrap

absidue:
As a dependency of apt gnutls needs to be in the bootstrap.


Diatrus:
What about gnutils dependencies? Lol
Don’t forget about the new gpgv dependencies too.

Diatrus:
Also that’s not how to add a strapproject entirely

PR #701, opened 2021-05-02
add cups

demhademha:
This is in alpha stage, (currently, only the core tools compile), but it seems to work. Any feedback would be appreciated

### DIFF
--- a/cups.mk
+++ b/cups.mk
@@ -1,0 +1,46 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+SUBPROJECTS   += cups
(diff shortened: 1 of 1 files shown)


asdfugil:
what applications can use it?

CRKatri:
iOS￼ and macOS can already print,so why￼ ￼do we need this?

m1stadev:
> iOS￼ and macOS can already print,so why￼ ￼do we need this?

CLI all the things

demhademha:
One advantage of cups is that it will allow printing directly from cli, another advantage is that it will allow printing to non AirPrint printers. It can also be used by x11

PR #706, opened 2021-05-03
add libiosexec

sunflsks:


Diatrus:
Can you make it libiosexec1 and have a libiosexec-dev package with the static lib as well as a symlink `libiosexec.dylib -> libiosexec.1.dylib`

PR #713, opened 2021-05-06
Copy libtool archive to build_base

brennanleblanc0:
Fixes an issue on Linux where building packages that use libiosexec throw an error saying it can't find the library


Diatrus:
This issue should be resolved in a different way. I think it will be resolved when a new version of libiosexec is released.

PR #721, opened 2021-05-07
add micro

sunflsks:
tested on darwin-amd64 and iphoneos-arm64


m1stadev (changes requested):
Add a check in `micro.mk` that checks for the `go` binary on the host system before attempting to build.

sunflsks:
Good catch, thank you

PR #722, opened 2021-05-07
Add txt2man

TheRealKeto:
This PR adds ``txt2man``, a CLI that converts flat ASCII text to man page format.


Diatrus:
Good shit as always

PR #725, opened 2021-05-09
Add cmark

absidue:
Markdown parsing and conversion tool and library. The Down library that Sileo uses is a wrapper around this.

The library doesn't have a SO version so the library package is called `libcmark0.29.0`, Debian does the same.

This has been tested on 64-bit macOS and iOS 14.


absidue:
Tested and working in M1 as well. Thanks capt

PR #733, opened 2021-05-11
fix slang

sunflsks:
I'm not sure if libslang fails to compile just for me, but if it does here is the fix


Diatrus:
slang should not be putting an etc folder in the sub prefix. Can you look into that?

sunflsks:
sure

PR #735, opened 2021-05-11
add libdvdnav4

asdfugil:


Diatrus:
Why is there a libdvdread4 if the soversion is 8?

asdfugil:
that is a mistake 
searched in debian sid and it should be 8
stable is outdated

asdfugil:
Igtm

CRKatri (changes requested):
The libdvdcss update breaks building, says that the DVD ioctls are not supported.

asdfugil:
building libdvdcss works on macOS
that's right after pulling from upstream and make clean too.

asdfugil:
`build_tools/compile.sh` works for libdvdnav, libdvdread and libdvdcss

PR #737, opened 2021-05-13
Add xonsh

TheRealKeto:
This PR adds ``xonsh``, a Python-powered shell language. Currently a draft, as this requires a specific POSIX compliant script in order to be added to ``/etc/shells``. 

In case you ever need to, compiling this is fairly easy. You'll need ``pip`` and [``python-prompt-toolkit``](https://github.com/prompt-toolkit/python-prompt-toolkit) installed globally.


TheRealKeto:
I also forgot to mention that this needs an update from Python, which removes the ``execv failed, trying shell`` warnings.

Diatrus:
Is there not a default python args?

TheRealKeto:
What are you referring to?

PR #739, opened 2021-05-13
login libiosexec

sunflsks:
This should fix eperm issues for non-MachO shells


sunflsks:
Tested on iphoneos to work for both binary and interpreted shells.

PR #741, opened 2021-05-14
add libbluray

asdfugil:
works fine on darwin-amd64


asdfugil:
i should make this not build for iOS

asdfugil:
all right this should be fine now

CRKatri:
Why is this disabled on iOS?

sunflsks:
Because it requires DiskArbitration.framework, which is only available on macOS.

CRKatri:
Oh ok.

TheRealKeto:
Rebase/merge with main again and move your Makefile to the ``makefile`` directory.

PR #758, opened 2021-05-20
add lua bitop

sunflsks:
This is required for building neovim. Also, I'm not 100 percent sure if this is the correct way to package a lua module, so if someone could check this out, that'd be great.


CRKatri:
Superseded by #973

PR #765, opened 2021-05-23
Fix system-cmds building for iOS 11

m1stadev:
Marking as a draft, because there's probably a better way to go about doing this. Builds fine for CF versions 1400 and 1700.


CRKatri:
The better way to do this is to use [apple_cmds](https://github.com/CRKatri/apple_cmds) which has makefiles for everything (although it only supports 12+) /s
But in all seriousness is looks fine to me.

CRKatri (approved):

Diatrus:
@m1stadev can you instead pull the latest commit and try? I think I fixed your issue.

PR #782, opened 2021-05-29
Fix macos suite names

asdfugil:


Diatrus:
I’m never going to have a suite below big sur, so denying.

PR #784, opened 2021-05-29
Update dri control to not force a specific version

Gymxo:
This always seems to mess stuff up :/


TheRealKeto:
You should also update the version of the Makefile, so that it appears as an upgrade on package managers.

Diatrus:
Not only is this wrong but dri isnt even used now.

Gymxo:
Alright

Gymxo:
Just trying to find out what’s wrong with dummy driver

PR #785, opened 2021-05-30
Add commonly-used Rust flags variable

TheRealKeto:
This PR adds a new variable, ``DEFAULT_RUST_FLAGS``, which provides commonly-used flags in Rust projects. Newly Rust project additions must use this flag upon merging moving forward.


Makefile
```diff
@@ -428,12 +428,16 @@ DEFAULT_SETUP_PY_ENV := \
 	CXXFLAGS="$(CXXFLAGS) -I$(BUILD_STAGE)/python3/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/include/$$(ls $(BUILD_STAGE)/python3/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/include)" \
 	CPPFLAGS="$(CPPFLAGS) -I$(BUILD_STAGE)/python3/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/include/$$(ls $(BUILD_STAGE)/python3/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/include)"
 
+DEFAULT_RUST_FLAGS := \
+	SDKROOT="$(TARGET_SYSROOT)" \
+	PKG_CONFIG="$(RUST_TARGET)-pkg-config"
```
m1stadev, line 440: Could the cargo build flags (`--release` and `--target`) be added to this as well, and then change the cargo call to this:
 ```mk
cargo build \
	$(DEFAULT_RUST_FLAGS)

m1stadev (changes requested):

Makefile
TheRealKeto: These flags have to be added beforehand. I have tested this before, and it did not work on my end. You can give it a shot and report back.

m1stadev: alright, never mind then.

TheRealKeto:
This PR is basically done; while I did agree with Adam that this would be merged before ``pueue``, that will not be the case anymore.. ``pueue`` needs more work to get the daemon working, and I don't want this PR to be halted by that.

PR #791, opened 2021-05-31
sudo: Update and hopeful fix for sudo_noexec.so

CRKatri:


Diatrus:
How exactly is this a fix?

CRKatri:
`sudo_noexec.so` should overwrite the `exec` family of functions, but because it's including `unistd.h` which includes `libiosexec.h`, only `ie_exec` functions will be overwritten, which won't work on programs that don't use `libiosexec`. So I define `LIBIOSEXEC_PRIVATE` before `libiosexec.h`'s inclusion, so that the real `exec` functions will be overwritten, and add extra replacement for the `ie_exec` functions for good measure.

Diatrus:
@CRKatri that's false. the header defines the exec functions as their ie_exec alternatives itself if you don't define `LIBIOSEXEC_INTERNAL`.

CRKatri:
What's false? Overwriting the real `exec` functions is more important than the `ie_exec` ones. Because the `ie_exec` ones aren't always used.

Diatrus:
The real exec functions are overwritten is what I'm saying lol

CRKatri:
No, without the patch, only the `ie_exec` functions are overwritten. The replacement by `libiosexec.h` is causing the overwriting for `exec` to only happen for `ie_exec`.

Diatrus:
Ahhh ok I see now. Thanks. In the future, all execs will be ie_exec.

CRKatri:
However we can't ensure that non Procursus bins/libs will use libiosexec so I can't imagine we'll be able to remove this patch.

DaOfficialRolex:
Please update this PR for 1.9.7p1. I checked out the PR and was able to confirm that sudo works as intended on macOS amd64 and iOS.

DaOfficialRolex (approved):

makefiles/sudo.mk
```diff
@@ -42,8 +42,11 @@ endif
 		--with-password-timeout=0 \
 		--with-passprompt="[sudo] password for %p: " \
 		sudo_cv___func__=yes \
+		ac_cv_have_working_snprintf=yes \
+		ac_cv_have_working_vsnprintf=yes \
 		$(SUDO_CONFIGURE_ARGS)
-	+$(MAKE) -C $(BUILD_WORK)/sudo
+	+$(MAKE) -C $(BUILD_WORK)/sudo \
+		SUDO_LDFLAGS='$(LDFLAGS) -static' # sudo passes a flag to libtool twice, one of which is passed to the linker
```
TheRealKeto, line 49: Do it like this instead.

```suggestion
		SUDO_LDFLAGS="$(LDFLAGS) -static" # sudo passes a flag to libtool twice, one of which is passed to the linker
```

makefiles/sudo.mk
```diff
@@ -42,8 +42,11 @@ endif
...
 	+$(MAKE) -C $(BUILD_WORK)/sudo install \
 		DESTDIR=$(BUILD_STAGE)/sudo \
 		INSTALL_OWNER=''
```
TheRealKeto, line 52: Fix this.

```suggestion
		INSTALL_OWNER=""
```

TheRealKeto (changes requested):
After these changes, this PR should be merged.

DaOfficialRolex:
@CRKatri I just wanted to let you know on macOS running the `sudo visudo` command ends up to the default `/etc/sudoers` instead of `/opt/procursus/etc/sudoers` and this with `/opt/procursus` being at the top of my path.

CRKatri:
Is `sudo` respecting you path? Or overwriting it, try `sudo ‘/opt/procursus/sbin/visudo` and let me know if it opens the right `sudoers`, if it does then I’ll look into getting `sudo` to add `/opt/procursus` to the path. You may also want to add `/opt/procursus` to your path in `/etc/{z,}profile` instead of `.zshrc` so that root will use procursus too.

DaOfficialRolex:
> 
> 
> Is `sudo` respecting you path or overwriting it? try `sudo /opt/procursus/sbin/visudo` and let me know if it opens the right `sudoers`, if it does then I’ll look into getting `sudo` to add `/opt/procursus` to the path. You may also want to add `/opt/procursus` to your path in `/etc/{z,}profile` instead of `.zshrc` so that root will use procursus too.

`sudo /opt/procursus/sbin/visudo` does open the correct sudoers. I will look into adding it to `/opt/procursus` as you said.

DaOfficialRolex:
>  You may also want to add /opt/procursus to your path in /etc/{z,}profile instead of .zshrc so that root will use procursus too.
@CRKatri To clarify I added  `export PATH="/opt/procursus/sbin:$PATH"` to both my .zshrc in my home user path and to `/etc/zprofile` which then correctly references visudo from procursus. Sorry for the late reply.

DaOfficialRolex:
Sudo has been update to 1.9.7p2. I have gotten to test it yet, but the changelog seems to reference no major changes and only various bug fixes. [Here is the changelog.](https://www.sudo.ws/stable.html#1.9.7p2) If you wish to see it for yourself.

DaOfficialRolex:
Can confirm this works as intended I did notice tho that on macOS this tend to write over the previous sudoers file. So whatever changes are made in that file once you update it seems to erase those changes. Not sure if this intended behavior so I thought it might be worth it to let you know.

CRKatri:
> Can confirm this works update as intended I did notice tho that on macOS this tend to write over the previous sudoers file. So whatever changes are made in that file once you update it seems to erase those changes. Not sure if this intended behavior so I thought it might be worth it to let you know.

That's because it is not listed as a conffile (which still isn't supported on Sileo, @elihwyma think you could look into that?).

DaOfficialRolex:
> 
> That's because it is not listed as a conffile (which still isn't supported on Sileo, @elihwyma think you could look into that?).

This issue happened on macOS as well to clarify in which I used APT to install the package. Not Sileo on macOS.

CRKatri:
> > That's because it is not listed as a conffile (which still isn't supported on Sileo, @elihwyma think you could look into that?).
> 
> This issue happened on macOS as well to clarify in which I used APT to install the package. Not Sileo on macOS.

It's not listed as a conffile, so APT will overwrite it, I mentioned Sileo, because sudo is part of the strap and we don't want to add a package to the bootstrap that throughs errors in Sileo.

PR #792, opened 2021-06-01
Update radare2

demhademha:
This pull request updates radare2 from version 4.5.0 to version 5.3.0


Diatrus:
This won’t pack a deb properly due to the ABI changing, and the indentation is off with the .PHONY again…

PR #794, opened 2021-06-02
Add xbattbar

asdfugil:
the battery-checking backend `xbattbar-check-darwin.m` is made by absidue.


build_patch/xbattbar/xbattbar-darwin.diff
```diff
@@ -0,0 +1,115 @@
...
++}
+diff -urN a/xbattbar.c b/xbattbar.c
+--- a/xbattbar.c	2021-06-02 22:00:04.000000000 +0800
++++ b/xbattbar.c	2021-06-02 22:09:12.000000000 +0800
+@@ -75,7 +75,7 @@
+ char *OFFIN_C  = "blue";
+ char *OFFOUT_C = "red";
+ 
+-char *EXTERNAL_CHECK = "/usr/lib/xbattbar/xbattbar-check-apm";
++char *EXTERNAL_CHECK = "/usr/lib/xbattbar/xbattbar-check-darwin";
```
Gymxo, line 112: Is that path actually hard coded?

Gymxo (changes requested):

build_patch/xbattbar/xbattbar-darwin.diff
asdfugil: it is, i guess i should use SED instead with $(MEMO_PREFIX)$(MEMO_SUB_PREFIX)

asdfugil:
`build_tools/compile_iphoneos.sh` seems to work fine

PR #795, opened 2021-06-02
Add tapi command

sunflsks:
Adds the `tapi` command and reorganizes package files.

If merged, #781 should probably be closed


CRKatri:
You may want to soversion the libtapi package, unless there's a reason @Diatrus didn't when he added it.

Diatrus:
@CRKatri i didn’t because it didn’t. Occasionally there is a package without an sover and that’s ok.

CRKatri:
It does though, `8svn` iirc, but I believe it just borrows it from the `llvm` that it compiles against. [`tapi-1200.0.23`](https://opensource.apple.com/release/developer-tools-120.html) may have an ABI change so it may be necessary to setup a so version in that case, but if Apple says "coming soon", who knows when that is.

Diatrus:
@CRKatri between versions 1000 and 1100 there was no abi change, so I expect the same here.

PR #798, opened 2021-06-06
Update less

TheRealKeto:
This PR fixes the control file for ``less``, which should provide pager. The pager can now be automatically installed alongside other packages (e.g Git) as long as ``pager`` is set as a dependency.


Diatrus:
> `Provides: most, more, pager`
less doesn't provide most.

PR #800, opened 2021-06-06
lf: Minor fixes and changes

TheRealKeto:
This PR updates ``lf`` to the latest commit hash, while also providing external files for the project, which were referenced from [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=lf), per request of Hayden.

The update also addresses a small issue with the displayed version number — refer to the example below. In short, compiling should now display the ``DEB_LF_V`` variable as the version (rather than ``LF_VERSION``), which matches the version used in the control file of the project.

```zsh
# Before
~ lf -version
r24

# After
~ lf -version
0~r24
```


makefiles/lf.mk
```diff
@@ -3,31 +3,37 @@ $(error Use the main Makefile)
...
 lf-setup: setup
 	$(call GITHUB_ARCHIVE,gokcehan,lf,$(LF_VERSION),$(LF_VERSION))
 	$(call EXTRACT_TAR,lf-$(LF_VERSION).tar.gz,lf-$(LF_VERSION),lf)
-	mkdir -p $(BUILD_STAGE)/lf/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/{bin,share/man/man1}
+	mkdir -p $(BUILD_STAGE)/lf/{etc/profile.d,$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/{bin,share/{lf,man/man1,zsh/site-functions,vim/vimfiles/syntax}}}
+
+ifneq (,$(findstring darwin,$(MEMO_TARGET)))
+LF_DEPS := lf-setup
+else
+LF_DEPS := lf-setup libiosexec
```
CRKatri, line 17: don't add libiosexec as a dep, it is useless.

TheRealKeto: It was added since everything was linked to it (through a huge commit); I'll remove it shortly.

makefiles/lf.mk
```diff
@@ -3,31 +3,31 @@ $(error Use the main Makefile)
 endif
 
 SUBPROJECTS += lf
-LF_VERSION  := r22
+LF_VERSION  := r24
```
CRKatri, line 6: Please update to `r26`

CRKatri (changes requested):

makefiles/lf.mk
TheRealKeto: Done. Please review again.

CRKatri (approved):

PR #801, opened 2021-06-06
Add shared library Makefile generator + control generator

asdfugil:


build_misc/templates/autotools-lib.mk
```diff
@@ -0,0 +1,54 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS   += @pkg@
+@PKG@_VERSION := @PKG_VERSION@
+@PKG@_SOVER   := @SOVER@
```
CRKatri, line 7: Don't use a variable for the SO version, this is only used when the package doesn't provide a SO version itself.

build_misc/templates/cmake-lib.mk
```diff
@@ -0,0 +1,54 @@
...
+@download@
+	$(call EXTRACT_TAR,@pkg@-$(@PKG@_VERSION).tar.@compression@,@pkg@-$(@PKG@_VERSION),@pkg@)
+	#$(call DO_PATCH,@pkg@,@pkg@,-p1)
+
+ifneq ($(wildcard $(BUILD_WORK)/@pkg@/.build_complete),)
+@pkg@:
+	@echo "Using previously built @pkg@."
+else
+@pkg@: @pkg@-setup
+	cd $(BUILD_WORK)/@pkg@ && ./configure -C \
```
CRKatri, line 20: This cmake template is using autotools.

build_misc/templates/cmake-lib.mk
```diff
@@ -0,0 +1,54 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS   += @pkg@
+@PKG@_VERSION := @PKG_VERSION@
+@PKG@_SOVER   := @SOVER@
```
CRKatri, line 7: Ditto

build_tools/new_library.sh
```diff
@@ -0,0 +1,92 @@
...
+	build="$(ask "Build System" $2)"
+	checkbuild "$build"
+	ver="$(ask "Package Version" $3)"
+	download="$(ask "Download Link" $4)"
+	sover="$(ask "SO Version" $5)"
+	description="$(ask "Description" "$6")"
+	extended_description="$(ask_extended_description "$7")"
+	echo "Writing out control and makefile"
+	createfromtemplate build_misc/templates/${build}-lib.mk ${pkg}.mk
+	createfromtemplate build_misc/templates/libpkg1.control build_info/${pkg}${sover}.control
```
CRKatri, line 81: Add the control file to `new_package.sh` and create one file that is sourced with the shared functions so that the two scripts are unified.

CRKatri (changes requested):

asdfugil:
i should probably add  ninja as well
but not python and perl because it doesn't make much sense of shared libraries

asdfugil:
most libraries doesn't need patching, so DO_PATCH is commented out

asdfugil:
All right

asdfugil:
Fixed makefile path

asdfugil:
no more libiosexec1 depends

build_misc/templates/pkg.control
```diff
@@ -0,0 +1,8 @@
+Package: @pkg@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Version: @DEB_@PKG@_V@
+Section: @Section@
+Priority: optional
+Description: @DEB_DESCRIPTION@
+@DEB_EXTENDED_DESCRIPTION@
```
TheRealKeto, line 8: Add a new line here.

build_tools/new_common.sh
```diff
@@ -0,0 +1,67 @@
...
+if [ -z "$1" ]; then
+	ESCAPED_DESCRIPTION=""
+	>&2 echo "Extended Description (press ENTER then Ctrl+D when done):"
+	while read line
+        	do ESCAPED_DESCRIPTION="$ESCAPED_DESCRIPTION"" $line\n"
+	done
+else
+	ESCAPED_DESCRIPTION="$(echo "$1" | ${SED} '$!s/$/\\n/' | tr -d '\n' | ${SED} 's|^| |g')"
+	fi
+	echo "$ESCAPED_DESCRIPTION"
```
TheRealKeto, line 47: Indent once.

TheRealKeto (changes requested):
For the templates, you can remove the DO_PATCH function; if needed, users can be pointed to the documentation as well as the wiki on how to apply patches.

build_misc/templates/pkg.control
asdfugil: @DEB_EXTENDED_DESCRIPTION@ already has a new line at end of file.

CRKatri:
This branch is very dirty. Also, imho this should be a part of the `new_package.sh` script, not a separate script.

PR #805, opened 2021-06-07
add hicolor-icon-theme

Gymxo:
Leaving the "Section:" blank until a contributor tells me where it should go
This is also a very important package for gtk stuff, As it's the main theme that's used.
This package has no depends but other things will depend on this.

### DIFF
--- a/build_info/hicolor-icon-theme.control
+++ b/build_info/hicolor-icon-theme.control
@@ -1,8 +1,8 @@
 Package: hicolor-icon-theme
 Version: @DEB_HICOLOR-ICON-THEME_V@
-Architecture: @DEB_ARCH@
+Architecture: all
 Maintainer: @DEB_MAINTAINER@
-Section:
+Section: Miscellaneous
 Priority: optional
 Description: default fallback theme for FreeDesktop.org icon themes
  This is the default fallback theme used by implementations of the
--- a/build_info/hicolor-icon-theme.control
+++ b/build_info/hicolor-icon-theme.control
@@ -1,0 +1,9 @@
+Package: hicolor-icon-theme
+Version: @DEB_HICOLOR-ICON-THEME_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section:
+Priority: optional
+Description: default fallback theme for FreeDesktop.org icon themes
+ This is the default fallback theme used by implementations of the
+ Freedesktop.org Icon Theme specification.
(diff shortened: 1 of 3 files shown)


build_info/hicolor-icon-theme.control
```diff
@@ -0,0 +1,9 @@
+Package: hicolor-icon-theme
+Version: @DEB_HICOLOR-ICON-THEME_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section:
```
CRKatri, line 5: Personally I’d add a `Miscellaneous` section, but I’m not the one to ask.

build_info/hicolor-icon-theme.control
```diff
@@ -0,0 +1,9 @@
+Package: hicolor-icon-theme
+Version: @DEB_HICOLOR-ICON-THEME_V@
+Architecture: @DEB_ARCH@
```
CRKatri, line 3: This should be `all`.

build_info/hicolor-icon-theme.prerm
```diff
@@ -0,0 +1,8 @@
+#! /bin/sh
+set -e
+
+if [ "$1" = "remove" ] || [ "$1" = "deconfigure" ] ; then
+  rm -f /usr/share/icons/hicolor/icon-theme.cache
```
CRKatri, line 5: Use `@MEMO_PREFIX@@MEMO_SUB_PREFIX@`, and add the postinst and triggers from Debian.

CRKatri (changes requested):

build_info/hicolor-icon-theme.prerm
Gymxo: We don’t have gtk yet to which is why I’m waiting on that. If you want

CRKatri: I wouldn’t wait, it won’t fail if missing gtk, so might as well add it so that we don’t need future changes down the line.

Gymxo: I’ll start working on getting gtk uploaded this week(without gir for now)

CRKatri: Don't worry about getting gtk added before adding the trigger and postinst.

Gymxo: Na I'm gonna add the trigger and postinst. It's just that I have nothing to do for basically the whole week.

Gymxo:
I opened this too early

PR #810, opened 2021-06-08
Add Fully Working Gitea and Tea

CRKatri:
I actually tested this one.

image


CRKatri:
How to use:
```
sudo -u _gitea -P -i
mkdir -p /var/lib/gitea/custom/conf
cp /usr/share/gitea/example.ini /var/lib/gitea/custom/conf/app.ini
chmod 640 /var/lib/gitea/custom/conf/app.ini
gitea
```
Then finish the setup at `localhost:3000`.

Diatrus (changes requested):
I would prefer using _gitea rather than gitea. The underscore is traditionally used for users that won't really be logging in.

Diatrus:
> I would prefer using _gitea rather than gitea. The underscore is traditionally used for users that won't really be logging in.

If this really is better, however, I don't care that much. Just lmk what you think.

Diatrus:
> How to use:
> 
> ```
> sudo -u gitea -P -i
> mkdir -p /var/lib/gitea/custom/conf
> cp /usr/share/gitea/example.ini /var/lib/gitea/custom/conf/app.ini
> chmod 640 /var/lib/gitea/custom/conf/app.ini
> gitea
> ```
> 
> Then finish the setup at `localhost:3000`.

You can do these steps in the postinst, don't you think?

CRKatri:
> > I would prefer using _gitea rather than gitea. The underscore is traditionally used for users that won't really be logging in.
> 
> 
> 
> If this really is better, however, I don't care that much. Just lmk what you think.

You do log in to this user, when cloning over ssh, gitea@address:repo.git is better than _gitea@address:repo.git. However on mine (https://git.cameronkatri.com) I use the git user. So I figured using gitea as a good middle ground cause some other package may want git.

> > How to use:
> 
> > 
> 
> > ```
> 
> > sudo -u gitea -P -i
> 
> > mkdir -p /var/lib/gitea/custom/conf
> 
> > cp /usr/share/gitea/example.ini /var/lib/gitea/custom/conf/app.ini
> 
> > chmod 640 /var/lib/gitea/custom/conf/app.ini
> 
> > gitea
> 
> > ```
> 
> > 
> 
> > Then finish the setup at `localhost:3000`.
> 
> 
> 
> You can do these steps in the postinst, don't you think?

I don't think that's necessary, gitea will create an app.ini at that path automatically when run if it doesn't exist, so that is only to allow you to use the sane defaults for Procursus. If you want to actually use gitea you'd definitely need to edit that config.

Diatrus:
> I don't think that's necessary, gitea will create an app.ini at that path automatically when run if it doesn't exist, so that is only to allow you to use the sane defaults for Procursus. If you want to actually use gitea you'd definitely need to edit that config.

For sane defaults (keyword default) I feel like there should be an "if ini doesn't exist, copy the example over" in the postinst. Not to mention making that directory under the gitea user in the postinst.

CRKatri:
> For sane defaults (keyword default) I feel like there should be an "if ini doesn't exist, copy the example over" in the postinst. Not to mention making that directory under the gitea user in the postinst.

However some people like to put the config at `/etc/gitea.ini` so, what I'm gonna do is, leave the example.ini, but then also copy it over to `/etc` and list it as a conffile, and add a Launchd daemon to start Gitea with the config from `/etc`

Diatrus:
> > For sane defaults (keyword default) I feel like there should be an "if ini doesn't exist, copy the example over" in the postinst. Not to mention making that directory under the gitea user in the postinst.
> 
> However some people like to put the config at `/etc/gitea.ini` so, what I'm gonna do is, leave the example.ini, but then also copy it over to `/etc` and list it as a conffile, and add a Launchd daemon to start Gitea with the config from `/etc`

That sounds like a plan. Remember to do a wrapper like with tor and shshd, so it doesn't get jetsam'd as easy.

CRKatri:
Hopefully launchd’s user key works, otherwise I may have to do some shenanigans with `su`.

CRKatri:
Ok, the launch daemon works great, there is now a global config (which has to be writable by the gitea user so I added a statoverride) and I added the `tea` cli tool. @Diatrus

m1stadev:
Shouldn't the `WorkingDirectory` key be set in the LaunchDaemon, rather than just using a wrapper to change into the directory then run gitea?

CRKatri:
It doesn't really matter cause the config specifies /var/lib/gitea anyway

build_info/gitea.control
```diff
@@ -0,0 +1,9 @@
+Package: gitea
+Version: @DEB_GITEA_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section: Data_Storage
+Priority: optional
+Depends: git, libpam2
+Pre-Depends: system-cmds (>= 854.40.2-2)
+Description: Git with a cup of tea, painless self-hosted git service
```
TheRealKeto, line 9: Add the gitea homepage above the Description field; do this for the macOS control file as well.

build_info/gitea.control
```diff
@@ -0,0 +1,9 @@
+Package: gitea
+Version: @DEB_GITEA_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section: Data_Storage
+Priority: optional
+Depends: git, libpam2
+Pre-Depends: system-cmds (>= 854.40.2-2)
```
TheRealKeto, line 8: While the order doesn't matter, it makes more sense to write Pre-Depends first followed by Depends.

TheRealKeto (changes requested):
You might want to consider adding a prerm script for this in order to unload the daemon once the package is removed from the device; everything else LGTM.

makefiles/gitea.mk
```diff
@@ -0,0 +1,57 @@
...
+GITEA_VERSION := 1.14.3
+DEB_GITEA_V   ?= $(GITEA_VERSION)
+
+gitea-setup: setup
+	-[ ! -f "$(BUILD_SOURCE)/gitea-$(GITEA_VERSION).tar.gz" ] && \
+		wget -q -nc -O$(BUILD_SOURCE)/gitea-$(GITEA_VERSION).tar.gz \
+			https://github.com/go-gitea/gitea/releases/download/v$(GITEA_VERSION)/gitea-src-$(GITEA_VERSION).tar.gz
+	-[ ! -f "$(BUILD_WORK)/gitea" ] && \
+		mkdir -p $(BUILD_WORK)/gitea && \
+			tar xf $(BUILD_SOURCE)/gitea-$(GITEA_VERSION).tar.gz -C $(BUILD_WORK)/gitea
```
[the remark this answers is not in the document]
CRKatri: The release src tarball contains the frontend files already compiled, this removes the build. dependency on nodejs that would exist otherwise. This tarball also doesn't have a prefix, hence I used tar directly instead of using `EXTRACT_TAR`.

PR #812, opened 2021-06-10
Updated git to 2.32.0

DaOfficialRolex:


git.mk
```diff
@@ -3,13 +3,15 @@ $(error Use the main Makefile)
...
-DEB_GIT_V   ?= $(GIT_VERSION)-1
+GIT_VERSION := 2.32.0
+DEB_GIT_V   ?= $(GIT_VERSION)
 
 GIT_ARGS += uname_S=Darwin \
 	HOST_CPU=$(GNU_HOST_TRIPLE) \
 	DESTDIR=$(BUILD_STAGE)/git \
 	MANDIR=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/share/man \
+	PYTHON_PATH=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/python3
+	ASCIIDOC_PATH=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/asciidoc
```
TheRealKeto, line 14: For the record, you need to add an \ after this arg.

DaOfficialRolex: Surprising it worked fine without the \, but I will add.

TheRealKeto (approved):
LGTM! :)

DaOfficialRolex:
This builds and works correctly. Please review before merging

makefiles/git.mk
```diff
@@ -3,13 +3,15 @@ $(error Use the main Makefile)
...
-DEB_GIT_V   ?= $(GIT_VERSION)-1
+GIT_VERSION := 2.32.0
+DEB_GIT_V   ?= $(GIT_VERSION)
 
 GIT_ARGS += uname_S=Darwin \
 	HOST_CPU=$(GNU_HOST_TRIPLE) \
 	DESTDIR=$(BUILD_STAGE)/git \
 	MANDIR=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/share/man \
+	PYTHON_PATH=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/python3 \
+	ASCIIDOC_PATH=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/bin/asciidoc \
```
Diatrus, line 14: For the two new lines here, is this wanting the host bins or the build bins?

DaOfficialRolex: It needs the bins that are on the host system. Cause it looks for those bins before it can build.

CRKatri: `PYTHON_PATH` is used to set the shebang on scripts.

Diatrus: @CRKatri and asciidoc?

Diatrus:
Please remove the patch file if there is no more patching done

makefiles/git.mk
CRKatri: `ASCIIDOC_PATH` does not exist, but it needs asciidoc to compile the documentation during build.

sunflsks: Not sure if I'm understanding this correctly but if Git needs asciidoc to build why is `MEMO_{SUB_}PREFIX` being used to find it? Shouldn't it be `$(which asiidoc)` or something similar?

DaOfficialRolex: `ASCIIDOC_PATH` does nothing now that I have rebuilt it twice. If you have ASIIDOC installed on the system it will be build the docs regardless of it being a variable. It will be removed now.

PR #818, opened 2021-06-14
zsh: Change installation path

TheRealKeto:
This PR introduces a minor change on where ``zsh-syntax-highlighting`` is installed; sourcing ``zsh-syntax-highlighting.zsh`` should be a lot shorter as opposed to older versions.

```zsh
# Before
$ source /usr/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh

# After
$ source /usr/share/zsh/zsh-syntax-highlighting.zsh
```

I also changed some minor mistakes in the control file and added a link to the homepage of the project.

### DIFF
--- a/build_info/zsh-syntax-highlighting.control
+++ b/build_info/zsh-syntax-highlighting.control
@@ -6,10 +6,11 @@ Depends: zsh
 Enhances: zsh
 Section: Terminal_Support
 Priority: optional
+Homepage: https://github.com/zsh-users/zsh-syntax-highlighting
 Description: Fish shell like syntax highlighting for zsh
- This package provides syntax highlighting for the shell zsh.  It enables
+ This package provides syntax highlighting for the shell zsh. It enables
  highlighting of commands whilst they are typed at a zsh prompt into an
- interactive terminal.  This helps in reviewing commands before running
+ interactive terminal. This helps in reviewing commands before running
  them, particularly in catching syntax errors.
  .
  This feature is inspired by the Fish shell, which provides it by default.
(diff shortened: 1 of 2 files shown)


CRKatri:
I don’t like this, the current way is how Debian does use.

TheRealKeto:
A compromise is to add both with a symlink; I don't see how this would affect "anything" that's **super** important when it comes to not following Debian, but whatever.

CRKatri:
I don't see the benefit to this. Also that isn't what /usr/share/zsh is for, it contains things that are sourced automatically iirc.

TheRealKeto:
This isn't targeting zsh, if you haven't noticed; This just provides a shorter path to ``zsh-syntax-highlighting``. Once again, a better solution would have been to just symlink the script to zsh, but since we need to follow how Debian does it, then I can change it.

Frankly, I don't see a reason not to add this; you haven't provided a convincing one; it's not like I've entirely renamed a package and it's not hard to look where files are installed...

Diatrus:
I don’t like doing this different than how everyone else handles it (e.g. Debain), because some people would copy their dotfiles over and need to change them to fit.

CRKatri:
1. Every package of this does it in it's own directory (Debian, Arch, FreeBSD, etc)
2. You didn't even do it for zsh-autosuggestions, so this would also create disunity within Procursus, not just between packages.

TheRealKeto:
Again, I can do this with a symlink, that way, the original composition of the package isn't strictly changed and other dotfiles don't just stop working; I cannot account for other packages I'm not aware of, so you can't blame that on me...

This is the best solution I can provide if "we" are strictly sticking to how Debian does it; Win-win.

CRKatri:
Or we could just leave it how it is, which is perfectly fine. You're shorter path is one line I type in a file and never think about again.

TheRealKeto:
Your issue is something that can be resolved, but whatever, Katri. The current path is the same as "mine" (because I can also do the same thing with the current path), but whatever. Please stop complaining about it now, Katri...

PR #826, opened 2021-06-18
Add binutils

asdfugil:


asdfugil:
the `binutils` package installs native tools prefixed with `g`

binutils.mk
```diff
@@ -0,0 +1,113 @@
...
+	# No architecure-specific control mechanism in PACK, sorry.
+	$(SED) -e "s/@TARGET@/$$($(SED) -e 's/_/-/g' <<< "$(GNU_HOST_TRIPLE)")/g" $(BUILD_INFO)/binutils-native.control.in > $(BUILD_INFO)/binutils.control
+	
+	# binutils.mk Sign
+	for target in $(BINUTILS_TARGETS); do \
+		$(call SIGN,binutils-$$target,general.xml);\
+	done
+	
+	# Make control files with:
+	# sed "s/@TARGET@/$(sed 's/_/-/g' <<< "$target")/g" binutils.control.in > binutils-$target.control
```
CRKatri, line 97: I know I added this line in just as a comment so adding control files is easy, but you should change it to generate the control files on the fly, simple to how you generate the native one.

CRKatri:
Does this build on macOS? And do all these BSD triples work? Cause when I did this the BSD targets would always fail for reasons I didn’t care enough about to figure out.

asdfugil:
this builds fine on macOS

CRKatri:
However, a lot of these are completely useless because Procursus' LLVM only supports x86, and arm. So the sparc, riscv, alpha, powerpc, etc targets could be removed.

Diatrus:
> However, a lot of these are completely useless because Procursus' LLVM only supports x86, and arm. So the sparc, riscv, alpha, powerpc, etc targets could be removed.

They can be there for the possible future.

asdfugil:
controls are now generated on the fly

makefiles/binutils.mk
```diff
@@ -1,12 +1,12 @@
-ifneq ($(PROCURSUS),1)
+fneq ($(PROCURSUS),1)
```
CRKatri, line 1: Missing an `i`.

asdfugil:
Actually, let me add some baremetal targets as well

CRKatri:
Could you add FreeBSD targets? Iirc it was the OpenBSD targets that were failing for me.

makefiles/binutils.mk
```diff
@@ -0,0 +1,116 @@
...
+	if [ -f $(BUILD_WORK)/binutils/$$target/.build_complete ]; then \
+		echo "Using previously built $$target binutils"; \
+	else \
+		mkdir -p $(BUILD_WORK)/binutils/$$target; \
+		cd $(BUILD_WORK)/binutils/$$target && ../configure -C \
+			--build=$$($(BUILD_MISC)/config.guess) \
+			--host=$(GNU_HOST_TRIPLE) \
+			--prefix=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX) \
+			--libdir=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/$(GNU_HOST_TRIPLE)/$$target/lib \
+			--includedir=$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/$(GNU_HOST_TRIPLE)/$$target/include \
```
CRKatri, line 53: You may want to do `MEMO_TARGET` here not `GNU_HOST_TRIPLE`.

makefiles/binutils.mk
```diff
@@ -0,0 +1,116 @@
...
+		rm -rf $(BUILD_DIST)/binutils-$$target; \
+		rm -rf $(BUILD_INFO)/binutils-$$($(SED) 's/_/-/g' <<< "$$target").control; \
+	done
+	rm -rf $(BUILD_DIST)/binutils{,-common}
+	mkdir -p $(BUILD_DIST)/binutils
+	mkdir -p $(BUILD_DIST)/binutils-common/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)
+	
+	#  binutils.mk Prep binutils-*
+	for target in $(BINUTILS_TARGETS); do \
+		cp -r $(BUILD_STAGE)/binutils/$$target $(BUILD_DIST)/binutils-$$($(SED) -e 's/_/-/g' <<< "$$target"); \
```
CRKatri, line 74: `cp -a` not `-r`.

makefiles/binutils.mk
```diff
@@ -0,0 +1,116 @@
...
+	done
+	for target in $(BINUTILS_TARGETS); do \
+		for pkg in $(BUILD_DIST)/binutils-$$target; do \
+			for man in $$pkg/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/share/man/man1/*; do \
+				$(LN) -sf binutils-$$(basename $$man .1 | rev | cut -d- -f1 | rev ).1.zst $$man; \
+			done; \
+			rm -rf $$pkg/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/share/{locale,info}; \
+		done; \
+	done
+	cp -r $(BUILD_STAGE)/binutils/x86_64-linux-gnu/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/share $(BUILD_DIST)/binutils-common/$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)
```
CRKatri, line 85: Ditto

CRKatri (changes requested):

makefiles/binutils.mk
asdfugil: thinking about `LLVM_TARGET`

CRKatri: However, `LLVM_TARGET` has the min version in it, which we don't want.

asdfugil: How about `RUST_TARGET`?
(I know this is not rust)

CRKatri: However, `RUST_TARGET` is shared between audioOS and TvOS, and iOS arm64 and iOS arm64e. `MEMO_TARGET` is the only one that isn't used for multiple targets.

asdfugil:
there's still some unidentified broken targets which will be removed later

asdfugil:
a total of 83 targets

asdfugil:
actually alpine is i486-linux-musl

CRKatri:
Could you remove all the broken targets?

asdfugil:
They are already remived

asdfugil:
In particular, all System V targets are removed.

asdfugil:
aight

asdfugil:
marking this as ready as everything can be installed at once now

asdfugil:
Fixed those file conflicts as well as removing the rev references.

asdfugil:
There are no outstanding issues in this PR so @cameren2k this is very much done

> [07:56] joe: loss when will nick finish binutils

asdfugil:
also about Nick Chan#0001: that will probably stay dead until May 2022

asdfugil:
Unless it is an iPhoneOS15.0.sdk moment 
```
../../gold/gold-threads.cc:288:13: error: expected expression                               : once_(PTHREAD_ONCE_INIT)                                                                      ^                                                                           /Users/nick/Procursus/build_base/iphoneos-arm64/1700/usr/include/pthread.h:213:27: note: expanded from macro 'PTHREAD_ONCE_INIT'                                                #define PTHREAD_ONCE_INIT {_PTHREAD_ONCE_SIG_init, {0}}                                                           ^
1 error generated.
```

asdfugil:
later

asdfugil:
this may actually have issues in building now, i need to test, also, the patch does not apply anymore on binutils > 2.38

PR #827, opened 2021-06-18
Prepare for new PojavLauncher update

Doregon:
We’ve added another Storyboard and switched to using Assets.xcassets with our new versions. 

This is a draft until we actually release the next version, or we end up pinging Hayden in our Discord asking him to update.

### DIFF
--- a/apt.mk
+++ b/apt.mk
@@ -3,7 +3,7 @@ $(error Use the main Makefile)
 endif
 
 STRAPPROJECTS += apt
-APT_VERSION   := 2.3.1
+APT_VERSION   := 2.3.2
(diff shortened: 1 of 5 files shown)


Diatrus:
Thank you. I’d rather not use xcrun, though, and run it directly. If you don’t fix that up I will soonish. The arguments there are also not friendly to Pro’s min ios ver stuff so we need to check on that.

Doregon:
@Diatrus Done! I switched to using the full path for `actool`, and use `IPHONEOS_DEPLOYMENT_TARGET` to comply with Proc's minimum version stuffs :P

Diatrus:
Thank you much. I'll do a proper review as soon as I'm home with a commit #

PR #833, opened 2021-06-23
add xmlto

sunflsks:


Diatrus:
Needs newer getopt or something

PR #842, opened 2021-06-25
all da targets

asdfugil:
- bridgeOS
- audioOS
- arm64e


Diatrus:
I don’t want audioOS or bridgeOS targets. They aren’t useful.

asdfugil:
bridgeOS I get it, but how about audioOS?

asdfugil:
you could very much use it a HomePod as a server.
It is practically always powered on and connected to internet.
and it has reasonable storage for that purpose

Diatrus:
> you could very much use it a HomePod as a server.
> It is practically always powered on and connected to internet.
> and it has reasonable storage for that purpose

I don't think it has nearly enough storage for a valid server. IIRC the storage couldn't even handle a full bootstrap.

asdfugil:
Well, it has 16 GB  storage and we can probably use up to 8 GB.

Diatrus:
That's alright to me, then.

PR #844, opened 2021-06-25
Add wiki to subfolder

CRKatri:
You'll need to do this: https://github.com/SwiftDocOrg/github-wiki-publish-action#3-generate-a-personal-access-token.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,16 @@
+name: Documentation
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Upload Documentation to Wiki
+        uses: SwiftDocOrg/github-wiki-publish-action@v1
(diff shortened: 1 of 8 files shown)


TheRealKeto:
This isn't the way these should be handled. Use a submodule.

CRKatri:
I don't think having it as a submodule is a good idea, mainly because submodules are a pain to update, either having it part of the repo or as a separate repo are better choices in my opinion.

TheRealKeto:
I don't think something like Theos' wiki handles the workflow within the Theos repository, so if a submodule isn't the way to go, then I see no point in adding them here.

Like I said previously, PRs for the Wiki will be handled in a new repository, so the workflow would go on that repository rather than here.

TheRealKeto:
Please wait for the repository to be done, and then we can move forward; give me time to get things ready rather than just jumping into it :)

TheRealKeto:
I think you can close this; the repo has been made and I've already made a pull request adding this change. You can find the repo by using the reference above.

PR #849, opened 2021-06-27
Re-add `rev` to coreutils

CRKatri:
In fb4187c, when `bsdcp` was removed, `rev` was no longer being
installed, then in 02a6b94, `rev` was no longer being downloaded or
built, which means currently iOS does not have `rev`.


Diatrus:
I’m ok with iOS not having rev right now. I’m going to add text-utils sometime. Not many people use rev anyways.

CRKatri:
I open a PR tomorrow for text_cmds.

PR #851, opened 2021-06-28
Add text_cmds

CRKatri:


Diatrus:
Why's this being put in strap?

CRKatri:
Because it has binaries I think to be essential, but I’ll remove it.

PR #862, opened 2021-06-30
Add ddrecovery

demhademha:


build_info/ddrescue.control
```diff
@@ -0,0 +1,45 @@
...
+ areas of the file, and for multiple recovery attempts over
+ different subsets.
+ .
+ gddrescue aligns its I/O buffer to the sector size so that it
+ can be used to read from raw devices. For efficiency reasons,
+ also aligns it to the memory page size if page size is a
+ multiple of sector size.
+ .
+ Please note that this is the GNU ddrescue version providing the
+ ddrescue executable. The package is named gddrescue because the
```
CRKatri, line 43: Either remove this paragraph or rename the package, I'd just remove the section.

makefiles/ddrescue.mk
```diff
@@ -0,0 +1,42 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS   += ddrescue
```
CRKatri, line 5: Alignment.

CRKatri (changes requested):

makefiles/ddrescue.mk
```diff
@@ -0,0 +1,42 @@
...
+DDRESCUE_VERSION := 1.25
+DEB_DDRESCUE_V   ?= $(DDRESCUE_VERSION)
+
+ddrescue-setup: setup
+	wget -q -nc -P $(BUILD_SOURCE) http://mirror.keystealth.org/gnu/ddrescue/ddrescue-$(DDRESCUE_VERSION).tar.lz
+	$(call EXTRACT_TAR,ddrescue-$(DDRESCUE_VERSION).tar.lz,ddrescue-$(DDRESCUE_VERSION),ddrescue)
```
TheRealKeto, line 11: Download the package signature provided in the mirror, and add a check to the Makefile with ``PGP_VERIFY``.

build_info/ddrescue.control
TheRealKeto: You should also change all references of the old package name in case you rename it.

build_info/ddrescue.control
```diff
@@ -0,0 +1,40 @@
+Package: ddrescue
```
TheRealKeto, line 1: I'd actually prefer this package name to be up with Debian (``gddrescue``), since I've had this issue where I can't figure what to install since packages we have have different names.

You should also rename your Makefile and control file to reflect this name as well.

build_info/ddrescue.control
```diff
@@ -0,0 +1,40 @@
...
+Version: @DEB_DDRESCUE_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section: Utilities
+Priority: optional
+Description: GNU data recovery tool
```
TheRealKeto, line 7: Add the [homepage](https://www.gnu.org/software/ddrescue/ddrescue.html) of the package to the Homepage field here.

TheRealKeto (changes requested):

build_info/ddrescue.control
CRKatri: Well considering (`ddrescue` was removed from Debian because of confusion between it and GNU `ddrescue`)[https://tracker.debian.org/news/444590/bug677101-removed-packages-from-unstable/] I think it’s safe to name to differently from Debian.

TheRealKeto: Most packages have always maintained their names based on what Debian names them so that there's no confusion, and we should try to avoid that to begin with.

demhademha:
All issues should now be resolved

CRKatri:
Does Debian ship with a g prefix on the bin out just on the package?

TheRealKeto:
> Does Debian ship with a g prefix on the bin out just on the package?

No, but that's on their end; we can have both ``ddrescue`` and ``gddrescue`` with a symlink.

demhademha:
If debian don't prefix each binary with a **g**, then I don't think we should either

demhademha:
Removed the old files.

PR #890, opened 2021-07-12
Add new_keyring.sh

CRKatri:


TheRealKeto:
Keyrings should be handled like tarballs; good job on the template. However, I'd add the Author field to the control file.

CRKatri:
Wdym like tarballs? If you mean download them, then no, that is insecure and currently all keyring packages have them tracked in git. Also, PGP keys don’t have authors.

TheRealKeto:
I don't know how it's insecure. We're not a server that stores PGP keys and revoke certificates; each individual who would like to have their keyring in the build system should make their package download the signature from a different source, just like we download a tarball from a mirror or Github. We don't need to keep track of these files ourselves!

Diatrus:
We should keep track of keys ourselves.

TheRealKeto (approved):
Add every project tarball, then. Let's not run the risk that those might be insecure too... zzz...

Diatrus:
There’s a big difference between a code tarball and a gpg key

PR #906, opened 2021-07-16
Fix #893: Correctly assign new PRs to PB

TheRealKeto:
This PR fixes the small workflow introduced in [#893](https://github.com/ProcursusTeam/Procursus/pull/893); the workflow no longer handles issues, since those can be added to the project board manually, and isn't a hassle. On top of that, not all issues filed apply to the project board.


TheRealKeto:
For the record, either someone who's part of the organization (and has push access to the repository) or Hayden himself must setup the required secret for the workflow to work.

PR #909, opened 2021-07-19
add gtime

demhademha:


build_info/gtime.control
```diff
@@ -0,0 +1,9 @@
+Package: gtime
+Version: @DEB_GTIME_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section: Utilities
+Priority: optional
+Homepage: https://www.gnu.org/software/time/
+Description: The `time' command runs another program, then displays information about the resources used by that program.
+
```
TheRealKeto, line 9:
```suggestion
```

build_info/gtime.control
```diff
@@ -0,0 +1,9 @@
+Package: gtime
+Version: @DEB_GTIME_V@
+Architecture: @DEB_ARCH@
+Maintainer: @DEB_MAINTAINER@
+Section: Utilities
+Priority: optional
+Homepage: https://www.gnu.org/software/time/
+Description: The `time' command runs another program, then displays information about the resources used by that program.
```
TheRealKeto, line 8: This one represents Debian's description more accurately.

```suggestion
Description: GNU time program for measuring CPU resource usage.
 The 'time' command runs another program, then displays information
 about the resources used by that program, collected by the system while
 the program was running. You can select which information is reported
 and the format in which it is shown, or have 'time' save the information
 in a file instead of display it on the screen. The resources that 'time'
 can report on fall into the general categories of time, memory, I/O, and
 IPC calls. The GNU version can format the output in arbitrary ways by 
 using a printf-style format string to include various resource measurements.
```

makefiles/gtime.mk
```diff
@@ -0,0 +1,46 @@
...
+ifneq ($(wildcard $(BUILD_WORK)/gtime/.build_complete),)
+gtime:
+	@echo "Using previously built gtime."
+else
+gtime: gtime-setup
+	cd $(BUILD_WORK)/gtime && ./configure -C \
+		$(DEFAULT_CONFIGURE_FLAGS) \
+		--program-prefix=g \
+		--with-packager=Procursus
+		--with-packager-bug-reports=https://github.com/ProcursusTeam/Procursus/issues
```
TheRealKeto, line 23: Just in case...

```suggestion
		--program-prefix="g" \
		--with-packager="Procursus" \
		--with-packager-bug-reports="https://github.com/ProcursusTeam/Procursus/issues"
```

TheRealKeto (changes requested):

PR #922, opened 2021-07-25
Force https, and use ftpmirror.gnu.org not ftp.gnu.org or a hardcoded mirror

CRKatri:


TheRealKeto (approved):

CRKatri:
Can someone please either get rid of or fix this workflow.

TheRealKeto:
The workflow works; it requires someone setting up the secrets — Hayden, or someone who's a member of the org (and has access to the preferences of the repository) can set that up.

I've talk to Hayden about using the project board within the repository (rather than the one outside), but the project board needs a specific setup. Once that's done, I'll PR the changes necessary for the workflow to work.

Diatrus:
We can’t force https without testing every one of these downloads you changed. I remember multiple downloads don’t have https on their sites.

TheRealKeto:
> We can’t force https without testing every one of these downloads you changed. I remember multiple downloads don’t have https on their sites.

If these are indeed mirrors of GNU, shouldn't they have all packages? I don't oppose testing, but I see no reason for any of these packages to fail.

CRKatri:
There are actually quite a few that don’t support https, like newosxbook and midnight commander

PR #925, opened 2021-07-28
Removed a ARM specific macro for amd64.

DaOfficialRolex:
There may be a better way to make this not build for amd64 systems, but I chose the path of least resistance. This macro is why amd64 systems were not able to build python and got the install autoconf-archive error even though it was installed.


Diatrus:
This isn’t at all the solution to what your problem was.

DaOfficialRolex:
I fail to see how, but please let me know if you find a better solution.

Diatrus:
Future reference: the issue was, he didn't have` automake` installed.

PR #944, opened 2021-08-07
Add cJSON

absidue:
This pull request adds the cJSON library because it's a dependency of the [mosquitto](https://mosquitto.org/) MQTT library and broker that I will add in a later pull request.


absidue:
Including the so version symlinks in the `libcjson1` package because clang links to those instead of the actual library.

PR #948, opened 2021-08-10
Very WIP krb5 debianizing.

DaOfficialRolex:
Will do this with whatever time I can scrap together.

### DIFF
--- a/makefiles/krb5.mk
+++ b/makefiles/krb5.mk
@@ -1,0 +1,81 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+SUBPROJECTS   += krb5
+KRB5_VERSION := 1.19.2
+DEB_KRB5_V   ?= $(KRB5_VERSION)
+
+krb5-setup: setup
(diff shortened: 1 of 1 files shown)


DaOfficialRolex:
Just need to add control files clean up this makefile and then it should be done.

DaOfficialRolex:
Stuck on this part for deb packaging. I am guessing the error is due to the fact the the deb has not be packed yet, but I am not sure.

```dpkg-deb: error: parsing file '/Users/rolex/Procursus/build_dist/darwin-amd64/1700/krb5-admin-server/DEBIAN/control' near line 6 package 'krb5-admin-server':
 'Version' field value '@DEB_KRB5_V@': version number does not start with digit
gmake[1]: *** [makefiles/krb5.mk:112: krb5-package] Error 2
gmake: *** [Makefile:1053: rebuild-krb5-package] Error 2```

DaOfficialRolex:
Done please review and merge when you have time.

CRKatri:
I fixed the `Depends:` on all of these, but multiple of them seem to link to `libverto.0.dylib`, which does not exist. I also noticed that the install_names contain the minor version too.

DaOfficialRolex:
TO DO: Fix actual Install_names possibly located in `~/krb5/src/config/libnover.in` and `~/krb5/src/config/lib.in`

PR #959, opened 2021-08-22
update radare2

### LINKED ISSUE #957: Package update: radare2
https://github.com/radareorg/radare2/archive/refs/tags/5.4.0.tar.gz

sunflsks:
gm
closes #957


CRKatri (changes requested):
Can you fix this and update it to 5.5.4? It does not seem to be respecting `CC`.

PR #960, opened 2021-08-25
debianize libirecovery

asdfugil:


Diatrus:
Get rid of the old libirecovry3 if you’re going to do this.. fix dependencies too.

PR #965, opened 2021-08-29
Initial iOS 15 + /private/preboot/procursus support

m1stadev:
Updating the base boostrap for initial iOS 15 support, which will requre the entire bootstrap to be inside of /private/preboot/procursus

TODO:
- [x] Fix system-cmds building
- [x] Fix sudo building
- [x] Fix base building
- [x] Disable firmware-sbin building
- [ ] General testing
- [x] Update libiosexec tbd


CRKatri (approved):

Makefile
```diff
@@ -124,6 +124,44 @@ ON_DEVICE_SDK_PATH   := $(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/share/SDKs/iPhoneOS.sdk
 BARE_PLATFORM        := iPhoneOS
 export IPHONEOS_DEPLOYMENT_TARGET
 
+else ifeq ($(MEMO_TARGET),iphoneos-arm64-preboot)
+ifneq ($(MEMO_QUIET),1)
+$(warning Building for iOS with preboot prefix)
+endif # ($(MEMO_QUIET),1)
+MEMO_ARCH            := arm64
+PLATFORM             := iphoneos
+DEB_ARCH             := iphoneos-arm
```
CRKatri, line 133: Change this to `iphoneos-arm64`

CRKatri (changes requested):

Makefile
m1stadev: Done.

asdfugil:
there's also watchOS, tvOS, bridgeOS and audioOS.

m1stadev:
That can be added once we know for a fact that SSV actually exists on the root volume on tvOS 15 (probably does, but should be sure, and iirc nobody's confirmed it yet)

asdfugil:
also needs iphoneos-arm migration

asdfugil:
Tweaks *should* work, accroding the situation with rootlessjb.

asdfugil:
- Automatically Move stuff to /p/p/p
  - Move stuff in /L/MS/DL to /p/p/p/lib/TweakInject
- Mach-O linked libraries header patch
  - Append /p/p/p, remove /usr for non-system libraries
  - Inject libiosexec
- Resign
- dpkg should print a warning

m1stadev:
We’re keeping `/usr`, so the path will be `/p/p/p/usr`

asdfugil:
Will we do usrmerge?

PR #974, opened 2021-08-31
Create mirrors.txt

CRKatri:
Having the mirror list on a server that can go down is not very discoverable.


m1stadev:
@TheRealKeto Could you add these mirrors to a page on the Procursus docs?

asdfugil (approved):
SLAM

PR #983, opened 2021-09-10
profile.d: set LANG and LANGUAGE accroding to system locale

asdfugil:


CRKatri (approved):

CRKatri:
Much more language than `zh_HK` will require an override, I'm going to close this, as users should set this themselves.

asdfugil:
NewTerm already set this,so I guess this is actually redundant for most users

PR #991, opened 2021-09-11
Remove useless workflow.

DaOfficialRolex:


CRKatri (approved):

asdfugil (approved):

TheRealKeto:
No. Hayden needs to setup the project board for this to work.

Unless there's a way to automate moving tiles from one card to another in project boards, I don't want this to be removed.

DaOfficialRolex:
The project board is not really used in any way and not even checked by most of the people who contribute to Procursus. If you really can't spend 5 mins to check it and edit it yourself. Then why would you create a workflow for it. Make a workflow that doesn't contribute to spamming my inbox with useless stuff.

TheRealKeto:
What your recieve on your inbox is not my issue.

Other alerts you might get from somewhere else are due to the fact that the project board isnt setup right, and the workflow hasn't been fixed because of that.

I guess automating a simple task isn't worth anyone's time because "my inbox gets spammed". Maybe set things up on your email correctly.

DaOfficialRolex:
> 
> 
> What your recieve on your inbox is not my issue.
> 
> Other alerts you might get from somewhere else are due to the fact that the project board isnt setup right, and the workflow hasn't been fixed because of that.
> 
> I guess automating a simple task isn't worth anyone's time because "my inbox gets spammed". Maybe set things up on your email correctly.

Forget my inbox getting spammed the workflows serve no purpose especially if you yourself the only person who wants to keep the workflow doesn't even check the project board it is going to. I made changes to it today earlier and it was legitimately not touched by you or anyone else. So why keep something that serves no purpose?

TheRealKeto:
I had reasons, and the only reason why it doesnt work with the current project board (that isnt in the repository) is because Hayden refuses to set the required token for it, and I'm not going to force someone that doesn't answer me for personal reasons.

Getting straight with anyone (currently) has just led to dead ends, so maybe removing it showcases that.

Rather than fixing the problem, you'd rather completely get rid of something that requires less than a minute to do right, but I'm done commenting on this.

TheRealKeto (approved):

PR #992, opened 2021-09-12
Add functions that builds and installs stuff

asdfugil:
The function themselves are done, but is not actually used yet.


Makefile
```diff
@@ -610,6 +627,59 @@ DO_PATCH    = cd $(BUILD_PATCH)/$(1); \
 		fi; \
 	done
 
+CONFIGURE_MAKE_INSTALL = cd $(BUILD_WORK)/$@; \
+	if ! [ $(1) ]; then \
+		./configure -C $(DEFAULT_CONFIGURE_FLAGS); \
+	else ./configure -C $(1); \
```
m1stadev, line 633: I think it'd be cleaner to include DEFAULT_CONFIGURE_FLAGS regardless of whether extra autoconf flags are passed, since they're still getting included in makefiles where extra flags are being passed. Same for all the other install functions

Makefile
```diff
@@ -610,6 +627,59 @@ DO_PATCH    = cd $(BUILD_PATCH)/$(1); \
...
+
+MESON_MESON_INSTALL = cd $(BUILD_WORK)/$@/build && meson \
+	--cross-file cross.txt\
+	.. $(1); \
+	$(2) DESTDIR='$(BUILD_STAGE)/$@' meson install $(3);
+
+PYTHON3_SETUP_PY_INSTALL = cd $(BUILD_WORK)/$@; \
+	if ! [ $(1) ]; then \
+		$(DEFAULT_SETUP_PY_ENV) python3 ./setup.py install \
+			--instal,l-layout=deb \
```
m1stadev, line 673: Typo

m1stadev (changes requested):

build_info/locale.sh
```diff
@@ -0,0 +1,9 @@
+SYSTEM_LOCALE=$(deviceinfo locale)
+if [ "$SYSTEM_LOCALE" != 'zh_HK' ]; then
```
CRKatri, line 2: Lots of other locales will need exceptions, not just `zh_HK`.

Makefile
asdfugil: *custom configure scripts* sigh

asdfugil:
I can't see this thing could be implemented in a sensible way

PR #1012, opened 2021-09-28
plutil: Use swift's open source plutil instead of bingner's

CRKatri:


CRKatri:
This `plutil` isn't quite good enough to be a replacement yet.

PR #1019, opened 2021-10-01
Update m4 to 1.4.19

Torrekie:


CRKatri:
Why is there a patch for pcre in here?

Torrekie:
sorry for this mistake, I'll move this to another PR later

PR #1023, opened 2021-10-04
Add yt-dlp: A youtube-dl fork

CRKatri:
The manpage is currently disabled because we do not have pandoc.


TheRealKeto:
youtube-dl also requires pandoc, yet its manpages get built regardless.

PR #1036, opened 2021-10-06
pam-modules: Install manpages

CRKatri:


makefiles/pam-modules.mk
```diff
@@ -42,16 +43,16 @@ endif
 pam-modules-package: pam-modules-stage
 	# pam-modules.mk Package Structure
 	rm -rf $(BUILD_DIST)/libpam-modules
-
+
```
m1stadev, line 46: Remove these extra spaces

m1stadev: fair enough then

PR #1038, opened 2021-10-07
Makefile: codesign with -o kill,hard,expires instead of -o runtime

CRKatri:
runtime includes library validation, which requires any libraries to be
system or signed with the same Team ID in order to be loaded, this means
that anything signed with a valid cert that is not part of Hayden Seay's
team will not work.

\> ldid
dyld: Library not loaded: @rpath/libcrypto.3.dylib
  Referenced from: /opt/procursus/bin/ldid
  Reason: no suitable image found.  Did find:
    /opt/procursus/lib/libcrypto.3.dylib: code signature in (/opt/procursus/lib/libcrypto.3.dylib) not valid for use in process using Library Validation: mapping process and mapped file (non-platform) have different Team IDs
    /opt/procursus/lib/libcrypto.3.dylib: stat() failed with errno=1
[1]    12218 abort      ldid

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -655,7 +655,7 @@ SIGN = 	CODESIGN_FLAGS="--sign $(CODESIGN_IDENTITY) --force --deep "; \
 		if [ "$(CODESIGN_IDENTITY)" != "-" ]; then \
 			CODESIGN_FLAGS+="--timestamp "; \
 			if [ "$(4)" != "nohardened" ]; then \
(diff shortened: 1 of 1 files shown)


Diatrus:
Sign with ldid adhoc too in this PR

sunflsks (changes requested):
remove the `expires` flag please

sunflsks (approved):

sunflsks (approved):

PR #1049, opened 2021-10-13
add jwm

asdfugil:
### All Submissions:

### Package Additions/Updates:


asdfugil:
`build_tools/compile.sh` works fine, also seems to be not prone to segfaults

Updated to 2.41

PR #1086, opened 2021-10-30
Rename GET_LOGICAL_CORES and make it configurable

1Conan:
- Not every CPU is using SMT.
- Make this configurable instead of forcing to use all cores/threads.
- Rename variable as it's not a function

Probably a better name than `CORE_COUNT` can be used though.

### All Submissions

* [ ] Have you made sure there aren't any other open [Pull Requests](https://github.com/ProcursusTeam/Procursus/pulls) for the same update/change?

### Package Additions/Updates

* [ ] Have you confirmed this builds & works as intended on an iOS device (if applicable)?


asdfugil:
also fixes the j flag on 1 thread or 3 threads systems

CRKatri (approved):

PR #1087, opened 2021-10-30
Update apt to 2.3.11

asdfugil:
### All Submissions

### Package Additions/Updates

### DIFF
--- a/build_patch/apt/undef-bool.diff
+++ b/build_patch/apt/undef-bool.diff
@@ -1,0 +1,13 @@
+diff -urN a/methods/basehttp.h b/methods/basehttp.h
+--- a/methods/basehttp.h	2021-10-21 18:21:53.000000000 +0800
++++ b/methods/basehttp.h	2021-10-30 21:21:50.528112471 +0800
+@@ -20,6 +20,9 @@
+ #include <string>
+ #include <time.h>
+ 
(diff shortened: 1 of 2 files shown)


build_patch/apt/undef-bool.diff
```diff
@@ -0,0 +1,13 @@
+diff -urN a/methods/basehttp.h b/methods/basehttp.h
+--- a/methods/basehttp.h	2021-10-21 18:21:53.000000000 +0800
++++ b/methods/basehttp.h	2021-10-30 21:21:50.528112471 +0800
+@@ -20,6 +20,9 @@
+ #include <string>
+ #include <time.h>
+ 
++#undef TRUE
++#undef FALSE
```
CRKatri, line 9: What's the point of this?

CRKatri (changes requested):

build_patch/apt/undef-bool.diff
asdfugil: to make the compile work :fr:

CRKatri: Yes but what's the error without it?

asdfugil: I think it is "expected identifier" in the enum following this

PR #1106, opened 2021-11-11
Add Kext tools

asdfugil:
### All Submissions

### Package Additions/Updates

 kextstat and mkextunpack works

kextstat and mkextunpack works


asdfugil:
- sys/cdefs.h are no longer overriden
- some patches made in here is not required when building on iOS and macOS, but needed when on Linux (didn't test FreeBSD)

asdfugil:
The iOS kernel appears to block kext from unloading, but does not appear to block kext loading.

asdfugil:
At least from [this](https://github.com/apple-oss-distributions/xnu/blob/main/libkern/c%2B%2B/OSKext.cpp) file anyways

CRKatri:
Why did you exclude `kextutil(8)`?

CRKatri (approved):

asdfugil:
It requires security.c, which requires some macOS only `Security.framework` function. Same goes for kextd.

asdfugil:
...and some implementation kextd is needed to load dexts.

PR #1136, opened 2022-01-06
Auto-generated DEBIAN/conffiles for all files in /etc

CRKatri:
build_info/*.conffiles should be reserved for files not in /etc that
need to be tracked in the conffiles.

We need to confirm Sileo, Zebra and Cydia will not choke on conffiles. Sileo and Zebra should work but the bundled versions in most jailbreaks is likely too old. Cydia will almost definitely need an update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -824,7 +824,7 @@ PACK = \
 			fi; \
 		done; \
 	done; \
-	(cd $(BUILD_DIST)/$(1)/; find .$(MEMO_PREFIX)/etc -type f) | cut -d. -f2- >> $(BUILD_DIST)/$(1)/DEBIAN/conffiles; \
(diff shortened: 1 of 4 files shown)


CRKatri:
Cydia on the repo has been update to not fail on packages with conffiles.
Zebra has been updated as well.
The version of Sileo bundled in Taurine and Odyssey has been updated to a version that doesn't die on conffiles.

The Cydia and Zebra updates have been available for a couple of weeks now.
I believe it's safe for this to be merged, but I would like approval from others before merging.

Diatrus (approved):

sunflsks (approved):

1Conan (approved):

m1stadev (approved):

PR #1142, opened 2022-01-31
use openssh for ramdisk

tihmstar:
Use openssh in ramdisk instead of dropbear, because dropbear is missing `scp`.

Tested on device.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1046,7 +1046,7 @@ env:
 
 include makefiles/*.mk
 
-RAMDISK_PROJECTS := bash coreutils diskdev-cmds dropbear findutils grep gzip ncurses profile.d readline tar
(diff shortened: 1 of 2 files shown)


CRKatri (approved):
Looks good, I will merge after I get a chance to test

badger200:
@tihmstar By sheer coincidence I just left you a detailed Issue on your ra1nsn0w GitHub asking about using it for an SSH ramdisk to hopefully recover data from my A9X iPad Pro with a dead internal 256GB NVMe NAND disk. If you have a chance sometime I'd be extremely grateful for your advice. I always used checkra1n with this device until the disk died, now it never even displays the chess logo, it just loops to recovery mode. Thanks again.

PR #1158, opened 2022-02-20
new package: xcodes

beerpiss:
### All Submissions

### Package Additions/Updates

----

Are macOS-only packages accepted?


CRKatri:
This tool depends on `aria2c`

beerpiss:
> This tool depends on `aria2c`

I wouldn't say "depends" since xcodes work fine without aria2c. I'll add aria2 as a suggested package.

From [RobotsAndPencils/xcodes](https://github.com/RobotsAndPencils/xcodes) README:
> If you have [aria2](https://aria2.github.io/) installed (it's available in Homebrew, brew install aria2), then xcodes will default to use it for downloads. It uses up to 16 connections to download Xcode 3-5x faster than URLSession.

CRKatri:
I think `Recommends` would be better than `Suggests`

PR #1159, opened 2022-02-20
Add pidof

asdfugil:
### All Submissions

### Package Additions/Updates

checksums from http://www.nightproductions.net/hashes/pidof_source.txt


CRKatri:
This does not belong in the bootstrap.

m1stadev:
> This does not belong in the bootstrap.

ratio + L bozo

m1stadev (approved):

makefiles/pidof.mk
```diff
@@ -0,0 +1,46 @@
+ifneq ($(PROCURSUS),1)
+$(error Use the main Makefile)
+endif
+
+STRAPPROJECTS += pidof
```
CRKatri, line 5: This is a `SUBPROJECT` not a `STRAPPROJECT`

CRKatri (changes requested):

asdfugil:
PROTIP: This has similar importance to `vi` if not more which also does not belong to the bootstrap so no one should ever try to add `vi` to the bootstrap.

CRKatri:
`ps aux | grep "process name"` seems kinda useless to me. It's also not on macOS or FreeBSD base installs, so I see no need for it to be part of the strap. It's importance is way less than `vi`, `vi` serves a legitimate purpose and deserves to be in the strap, but is not to keep the size down.
This is not going in the strap. Period.

makefiles/pidof.mk
asdfugil: All right

asdfugil:
It's useful for `renice(1)` but ok.

PR #1176, opened 2022-02-25
Update lrzip-next

demhademha:


CRKatri (changes requested):
`lrunzip` is segmentation faulting on M1.

demhademha:
Lrzuntar does not work with wildcards

demhademha:
Lrunzip as well

PR #1178, opened 2022-02-25
Update zlib-ng and make some modifications

demhademha:
As this is a drop in replacement for zlib, perhaps we should be replacing all non-system binaries with this zlib-ng


CRKatri:
No, we will continue to link the system zlib, linking zlib-ng is up to the person compiling.

CRKatri (approved):
Thanks.

PR #1191, opened 2022-02-28
libepoxy: fix wrong GL library paths

Torrekie:
### Description

libepoxy attempts to load `/opt/X11/lib/libGL.1.dylib` and OpenGL.framework which was not commonly exists on non macOS Darwins.

Patch it to correctly use Mesa.
### All Submissions

### Package Additions/Updates

### DIFF
--- a/build_patch/libepoxy/use-mesa.diff
+++ b/build_patch/libepoxy/use-mesa.diff
@@ -1,0 +1,57 @@
+diff -ur a/src/dispatch_common.c b/src/dispatch_common.c
+--- a/src/dispatch_common.c	2021-08-14 21:55:37.850696000 +0800
++++ b/src/dispatch_common.c	2022-02-28 15:46:08.729969790 +0800
+@@ -167,17 +167,24 @@
+ #include <err.h>
+ #include <pthread.h>
+ #endif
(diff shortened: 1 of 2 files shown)


build_patch/libepoxy/use-mesa.diff
```diff
@@ -0,0 +1,57 @@
...
+ 
+ #include "dispatch_common.h"
+ 
+-#if defined(__APPLE__)
++#if TARGET_OS_OSX
+ #define GLX_LIB "/opt/X11/lib/libGL.1.dylib"
+ #define OPENGL_LIB "/System/Library/Frameworks/OpenGL.framework/Versions/Current/OpenGL"
+ #define GLES1_LIB "libGLESv1_CM.so"
+ #define GLES2_LIB "libGLESv2.so"
++#elif TARGET_OS_EMBEDDED
```
Diatrus, line 23: This should be the case for all OSes here, not just embedded.

Torrekie: possibly not, still need to separate MAC and EMBEDDED and macOS have to do some further separation (/opt/X11 and /usr/lib/X11) but I don't have any older macOS to test

Issue #1171, opened 2022-02-24
m4: Segmentation fault in posix_spawn


asdfugil:
I ran into a segmentation fault when trying to do ` make shell-cmds` on iPad Pro 9.7 WiFi, iOS 14.3. It happens with m4.

m4 version: 1.4.19
libiosexec version: 1.0.20~1.1-alpha1

Debugger output:
```
Process 59809 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x90)
    frame #0: 0x00000001b40eb37c libsystem_kernel.dylib`posix_spawn + 156
libsystem_kernel.dylib`posix_spawn:
->  0x1b40eb37c <+156>: ldr    x9, [x8, #0x88]
    0x1b40eb380 <+160>: cbz    x9, 0x1b40eb3c4           ; <+228>
    0x1b40eb384 <+164>: ldrsw  x11, [x9, #0x4]
    0x1b40eb388 <+168>: mov    w10, #0x18
Target 0: (m4) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x90)
  * frame #0: 0x00000001b40eb37c libsystem_kernel.dylib`posix_spawn + 156
    frame #1: 0x00000001045d9994 libiosexec.1.dylib`ie_posix_spawn + 48
    frame #2: 0x00000001045d9b88 libiosexec.1.dylib`ie_posix_spawnp + 372
    frame #3: 0x0000000104594030 m4`___lldb_unnamed_symbol207$$m4 + 1012
    frame #4: 0x000000010457fd44 m4`___lldb_unnamed_symbol23$$m4 + 292
    frame #5: 0x00000001045881e4 m4`___lldb_unnamed_symbol85$$m4 + 1632
    frame #6: 0x0000000104587f94 m4`___lldb_unnamed_symbol85$$m4 + 1040
    frame #7: 0x0000000104587b40 m4`expand_input + 140
    frame #8: 0x000000010457e1d0 m4`___lldb_unnamed_symbol3$$m4 + 128
    frame #9: 0x000000010457dd38 m4`main + 1592
    frame #10: 0x0000000188633568 libdyld.dylib`start + 4
(lldb) c
```
...which might have something to do with libiosexec

Torrekie:
I didn't see m4 was using `posix_spawn` on my Apple M1 build
```
[...]

PR #1196, opened 2022-03-04
m4: fix segfaults and correct value detections

Torrekie:
### Description

Correct all guessed values, add vendor info, and fix issue #1171  

### All Submissions

### Package Additions/Updates


CRKatri:
I feel like most of those configure options can be discovered automatically, seems like a bad idea to hardcode all of those.

Torrekie:
some of them were important though but I just handle them all

CRKatri:
Superseded by f14be32

PR #1198, opened 2022-03-05
add vigr(8) and chkgrp(8)

Torrekie:
### All Submissions

### Package Additions/Updates


CRKatri:
These are not part of Apple's system-cmds, so I don't see the need to add them, this also puts the burden of maintaining them on us.

PR #1209, opened 2022-03-13
Makefile: add -Wl,-not_for_dyld_shared_cache to LDFLAGS

asdfugil:
-not_for_dyld_shared_cache is a ld option to mark a dynamic library that will
not got into dyld shared cache. No Procursus dylibs should go into dyld shared
cache, so we set it here. Setting this flag allows -undefined dynamic_lookup to
be used for dylibs that have an install name prefixed with /usr/lib. With this
flag, the linker will also skip adding additional information to the dylibs
generated, so we can save a little bit of disk space here. This flag is allowed
even with main executables even though it does not do anything.

Since LDFLAGS is actually passed to CC, we do -Wl,-not_for_dyld_shared_cache
here.

### All Submissions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -496,7 +496,7 @@ endif
 CFLAGS              := $(OPTIMIZATION_FLAGS) -arch $(MEMO_ARCH) -isysroot $(TARGET_SYSROOT) $(PLATFORM_VERSION_MIN) -isystem $(BUILD_BASE)$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/include -isystem $(BUILD_BASE)$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)$(MEMO_ALT_PREFIX)/include -F$(BUILD_BASE)$(MEMO_PREFIX)/System/Library/Frameworks -F$(BUILD_BASE)$(MEMO_PREFIX)/Library/Frameworks
 CXXFLAGS            := $(CFLAGS)
 CPPFLAGS            := -arch $(MEMO_ARCH) $(PLATFORM_VERSION_MIN) -isysroot $(TARGET_SYSROOT) -isystem $(BUILD_BASE)$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)/include -isystem $(BUILD_BASE)$(MEMO_PREFIX)$(MEMO_SUB_PREFIX)$(MEMO_ALT_PREFIX)/include -Wno-error-implicit-function-declaration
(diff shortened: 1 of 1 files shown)


Torrekie:
I just think we should remove that `-undefined dynamic_lookup`, shouldn't allow regular executables and libraries built with any wild undefined symbols

asdfugil:
Oh yeah we definitely should avoid using `-undefined dynamic_lookup`, but let's don't ignore the fact that it's used *everywhere* (well a lot of places I mean) , plus a warning will still be emitted with this flag.
also gets rid of this:
```
ld: warning: invalid -install_name (@rpath/libiosexec.1.dylib) in dependent dylib (/Users/foo/Documents/Procursus/build_base/iphoneos-arm64/1700/usr/lib/libiosexec.tbd). Dylibs/frameworks which might go in dyld shared cache cannot link with dylib that uses @rpath, @loader_path, etc.
```

Torrekie:
Normally only bundles (`*.so` or `*.bundle`) use dynamic lookup, they don't have install names so won't trigger that shared cache warning.

For modern autotools and cmake and other hell processes, they would not add flags that allows undefined symbols to be built while compiling dylibs and executables, authors do mark every required libraries and added their link flags, ignoring undefined symbols would be a bad idea under that situation

Torrekie:
And `-undefined dynamic_lookup` can and can only search symbols in loaded images, it will still throw errors if some undefined symbols was not in a library that in loaded list

A Hello World f90 program compiled without linking `libgfortran.dylib` but set `-undefined dynamic_lookup`
```
192:Procursus torrekie$ /opt/homebrew/Cellar/gcc/11.2.0_3/libexec/gcc/aarch64-apple-darwin20/11/collect2 -syslibroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -dynamic -arch arm64 -macos_version_min 11.0 -weak_reference_mismatches non-weak -o hello hello.o -undefined dynamic_lookup -lSystem
192:Procursus torrekie$ ./hello 
dyld: lazy symbol binding failed: Symbol not found: __gfortran_set_args
  Referenced from: /Users/torrekie/proj/Procursus/./hello
  Expected in: flat namespace

dyld: Symbol not found: __gfortran_set_args
  Referenced from: /Users/torrekie/proj/Procursus/./hello
  Expected in: flat namespace

Abort trap: 6
192:Procursus torrekie$ DYLD_INSERT_LIBRARIES=/opt/homebrew/Cellar/gcc/11.2.0_3/lib/gcc/11/libgfortran.dylib ./hello 
 Hello, World!
```

asdfugil:
This PR does not add `-undefined dynamic_lookup` to LDFLAGS, so that's a non-issue here.

asdfugil:
This means that when `-undefined dynamic_lookup` appears, the software author has adding them intentionally.

asdfugil:
This PR isn't even about `-undefined dynamic_lookup`...

Torrekie:
nah, but you would not need to add -not_for_dyld_shared_cache if there's no -undefined dynamic_lookup

CRKatri:
@Torrekie To be honest, I have no idea what you're going on about, it is not at all related to this PR.

CRKatri (approved):

CRKatri:
Thanks for the PR @asdfugil, sorry it took so long to merge.

PR #1248, opened 2022-05-23
apt: update to 2.5

asdfugil:
### All Submissions

### Package Additions/Updates

builds fine for iphoneos-arm64/1{5..7}00 and darwin-a{md,rm}64/1700.
Appears to install and work fine for iphoneos-arm64/1700 and darwin-amd64/1700 (actually 1800 with 1700 bootstrap)

iphoneos-arm64 note: you may encounter issues with libiosexec-dev dependencies. If you do, please build and install libiosexec debs from the GitHub repo.

### DIFF
--- a/build_patch/apt/path.diff
+++ b/build_patch/apt/path.diff
@@ -6,7 +6,7 @@ diff -urN apt-2.2.0/apt-pkg/init.cc apt/apt-pkg/init.cc
 
 
 -   Cnf.CndSet("DPkg::Path", "/usr/sbin:/usr/bin:/sbin:/bin");
-+   Cnf.CndSet("DPkg::Path", "/opt/procursus/sbin:/opt/procursus/bin:/usr/sbin:/usr/bin:/sbin:/bin");
++   Cnf.CndSet("DPkg::Path", "@MEMO_PREFIX@@MEMO_SUB_PREFIX@/sbin:@MEMO_PREFIX@@MEMO_SUB_PREFIX@/bin:@MEMO_PREFIX@/sbin:@MEMO_PREFIX@/bin:/usr/sbin:/usr/bin:/sbin:/bin");
 
     // Read an alternate config file
     _error->PushToStack();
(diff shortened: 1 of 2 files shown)


build_patch/apt-macos/path.diff
```diff
@@ -6,7 +6,7 @@ diff -urN apt-2.2.0/apt-pkg/init.cc apt/apt-pkg/init.cc
 
 
 -   Cnf.CndSet("DPkg::Path", "/usr/sbin:/usr/bin:/sbin:/bin");
-+   Cnf.CndSet("DPkg::Path", "/opt/procursus/sbin:/opt/procursus/bin:/usr/sbin:/usr/bin:/sbin:/bin");
++   Cnf.CndSet("DPkg::Path", "@MEMO_PREFIX@@MEMO_SUB_PREFIX@/sbin:@MEMO_PREFIX@@MEMO_SUB_PREFIX@/bin:/usr/sbin:/usr/bin:/sbin:/bin");
```
DaOfficialRolex, line 9: I don't think you need to change this at all since /opt/procursus will most likely always be the install path for Procursus on macOS. Also this patch will only compile for macOS so I really see no point in adding the variables here when this path will probably never change unless macOS significantly changes in the future

asdfugil: Would it be better if we keep thid patch like this, then unconditionally compile it?

DaOfficialRolex: Actually now I think about maybe this method should be consider cause of rootless stuff, but I will let @CRKatri decide that since he would know better

CRKatri: Move this patch to `build_patch/apt/` and keep the `@MEMO_PREFIX@@MEMO_SUB_PREFIX@` stuff, but don't forget to add `@MEMO_PREFIX@/bin` and `@MEMO_PREFIX@/sbin`.

CRKatri (approved):
Thank you!

PR #1005, opened 2021-09-23
Add dbus

Gymxo:

PR #1252, opened 2022-06-06
Makefile: Install expat.pc during setup on CFVER 1700+

asdfugil:
### All Submissions

makes #1005 buildable for CFVER 1700+


asdfugil:
the expat version actually shipped with iOS depends on the iOS version, so this is not the most correct PR
...but zlib and libxml is also like that, so maybe it is fine?
or should zlib, libxml and expat all be changed to match the system version in the Makefile?

PR #1254, opened 2022-06-09
Make shshd non-essential

L1ghtmann:
### All Submissions

* [x] This Pull Request doesn't contain any package additions; it's a small change (e.g., README change)

### Package Additions/Updates

* [ ] Have you confirmed this builds & works as intended on an iOS device (if applicable)?

### Preface
I believe I changed all of the relevant lines, but apologies in advance if I missed something.

### Thoughts
Despite shshd being an extremely useful package, I don't believe it should be marked as essential and installed with the strap. 

Most users likely don't care for or know what blobs are, so it makes very little sense to have a daemon uploading blobs for their device to a (trusted) 3rd party server without their knowledge or consent. If they know what blobs are and wish to save them, they'll see the value of this package and install it themselves...there's no need to do it for them. 

Moreover, marking it as essential and making it a dependency of the `essential` meta package (i.e., preventing it from being removed) is somewhat of an odd choice. Its purpose is not integral to the system and it cannot be easily disabled by a regular user (terminal = scary and launchctl = ??).


SerenaKit:
> Most users likely don't care for or know what blobs are, so it makes very little sense to have a daemon uploading blobs for their device to a (trusted) 3rd party server without their knowledge or consent 

While I agree the user consent part is a bit questionable, I do think it's very useful for users to have blobs either way, as, later, they may gain understanding of them, and in certain cases, the blobs uploaded by shshd may have saved them from losing their jb

CRKatri:
I agree we should remove it from essential, but keep it in the bootstrap. I'll merge this after you make those changes.

L1ghtmann:
@CRKatri Done. Appreciate your consideration! This seems like a good comprise.

CRKatri (approved):

PR #1276, opened 2022-08-27
Automatically insert libiosexec1 depend in PACK

### LINKED ISSUE #1274: Dangerously broken dependencies in package "APT"
I upgraded apt to 2.5.2 today using Sileo (to whom I will transfer the issue to if it turns out the problem isn't here). Apt gives the following error now:

`An unknown error occurred. APT was unable to find this package. Please try refreshing your sources dyld: Symbol not found: _ie_getgrnam Referenced from: /usr/lib/libapt- pkg.6.0.dylib Expected in: /usr/lib/libiosexec.1.dylib`

I poked around on my device, and found that the libiosexec is on an older version. (1.0.20) while the latest is 1.2.2. I am assuming the older version does not have that symbol.

The error, I am guessing, is in the dependencies of apt. The minimum libiosexec version needs to be bumped. To what? I am not sure, as Sileo does not show and apt is, haha, broken. Someone else needs to work out in which version that symbol was introduced.

Isn't there a GitHub CI which tests with the minimum versions specified? I think that should prevent similar events in the future. I am unfamiliar with CI, so I couldn't say how practical that would be.

Thankfully, dpkg seems to work. I will manually upgrade libiosexec using dpkg to work myself out of this pothole.

asdfugil:
### All Submissions

Fixes #1274
This PR makes it so that when all of the following conditions are met, a `libiosexec1` depend is auto-inserted into the control file.
- `darwin` is not in `$(MEMO_TARGET)`
- The package name is not `libiosexec1` or `libiosexec-dev`
- The package contains at least one Mach-O binary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -898,6 +898,17 @@ PACK = \
 			fi; \
 		done; \
 	done; \
+	if ! grep -q "darwin" <<< "$(MEMO_TARGET)"; then \
+		if ! [ "$(1)" = "libiosexec1" ] || ! [ "$(1)" = "libiosexec-dev" ]; then \
+			if find $(BUILD_DIST)/$(1) | xargs file -ib | grep 'x-mach-binary' | head -1 | grep -q 'x-mach-binary'; then \
+				if grep -q '^Depends\:' $(BUILD_DIST)/$(1)/DEBIAN/control; then \
+					sed -i 's/^Depends\:/Depends: libiosexec1 (>= '$${DEB_LIBIOSEXEC_V}'),/' $(BUILD_DIST)/$(1)/DEBIAN/control; \
+				else \
+					echo 'Depends: libiosexec1 (>= '$${DEB_LIBIOSEXEC_V}')' >> $(BUILD_DIST)/$(1)/DEBIAN/control; \
+				fi; \
+			fi; \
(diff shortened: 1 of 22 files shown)


Makefile
```diff
@@ -898,6 +898,17 @@ PACK = \
 			fi; \
 		done; \
 	done; \
+	if ! grep -q "darwin" <<< "$(MEMO_TARGET)"; then \
+		if ! [ "$(1)" = "libiosexec1" ] || ! [ "$(1)" = "libiosexec-dev" ]; then \
+			if find $(BUILD_DIST)/$(1) | xargs file -ib | grep 'x-mach-binary' | head -1 | grep -q 'x-mach-binary'; then \
```
CRKatri, line 903: Couldn't this be simplified to
```suggestion
			if find $(BUILD_DIST)/$(1) | xargs file -ib | grep -q 'x-mach-binary'; then \
```
Unless there was a reason for having the extra `head` and `grep` that I missed.

asdfugil: It could, but it won't stop on the first Mach-O if changed to this, and will take more time as a result.

CRKatri (approved):
LGTM!

PR #1300, opened 2022-11-05
hexedit: Update to 1.6

TheRealKeto:
This PR updates `hexedit` to its latest version, 1.6.


CRKatri (approved):
Thank you

PR #1315, opened 2022-11-23
libunistring: Update to 1.1

TheRealKeto:
This PR updates `libunistring` to its latest version, 1.1. A new package, `libunistring5` is introduced in this PR, which is a update replacement for `libunistring2`. Tested on iOS 12, building on macOS.


CRKatri (approved):
Thanks

PR #1322, opened 2022-11-30
Add VLC media player, usable as an iOS command line audio player

asdfugil:
This PR adds VLC media player, since I am not aware of a command line iOS audio player that can play many files formats.
On macOS, an working GUI `.app` is made

What works on iOS:
- Audio playback
- cursus TUI `vlc -I ncurses`
- CLI prompt UI `vlc -I oldrc`
- CLI, no UI `cvlc`

On macOS, it would install `VLC.app` and does work correctly

The following projects are added as build dependencies for VLC
- libaacs
- libdvdread
- libdvdnav
- libbdplus
- libbluray
- libmpeg2
- libnfs
- libsmb2

While making this PR, problems are discovered and fixed in the following packages:
- sdl2
- aom
- mpg123
- libcaca

`nasm` does not accept `CFLAGS`-like flags, so set `ASFLAGS` to empty when assembling x86.
misc: Verify libsodium download with a GPG key

### All Submissions

### Package Additions/Updates


m1stadev:
Would it be possible to add [VLC-iOS](https://github.com/videolan/vlc-ios) to this PR?

asdfugil:
just download it from the App Store...

asdfugil:
VLC-iOS is essentially, a wrapper of a wrapper around VLC for iOS and tvOS. Given that it is specifically made for iOS, it is must be somehow possible but I cannot see a reason to do that when it would be much better to download it from the App Store.

m1stadev:
Fair enough.

asdfugil:
this doesn’t seem to work on iphoneos-arm64/1800, will need fix it first

PR #1338, opened 2022-12-23
Update rar to 6.12

asdfugil:
- Fix rar for iOS 12/13 (tested on 12)
- Fix versioning scheme

### All Submissions

### Package Additions/Updates


CRKatri:
No thank you. Don't do weird hacks like that, iOS 12 and 13 support is irrelevant anyways.

Torrekie:
met same problem when truncating other macOS arm64 executables, seems caused by some different decisions between llvm and apple on whether keeping/using `___chkstk_darwin` in program. would be a hard requirement if going to port more macOS executables to iOS

PR #1342, opened 2023-01-13
hotfix: Fix priorities in (extreme-)clean

TheRealKeto:
Currently, directories such as `build_base` are cleared out when running `make clean`, which is not the intention of that Makefile rule. Instead, `clean` and `extreme-clean` have been restructured so that their names are not misleading.

This PR does not include any packages, so no checklist, Katri.


CRKatri (approved):
Sure whatever

PR #1404, opened 2023-06-12
Makefile: Delete .la symlinks when packing

TheRealKeto:
Several packages, e.g `libpng16` have several `*.la` files, some which are symbolic links. When packing, their parent files are deleted. In reality, parent files should be deleted alongside symbolic links pointing to them.


Teutates:
Great find.

